PR #7, opened 2015-05-25
Frist pass at docs for TDSCatalog.py.

lesserwhirls:
Also removed catalog_util.py - everything can be done by TDSCatalog

Note that the vcr based test (test_dods.py) is failing, but not sure why. Maybe because I replaced the call to catalog_util.get_latest_dods_url with a call to TDScatalog.get_latest_access_url?


dopplershift:
Docs look good.

One thing that pops out at me (unrelated) as I look at the code is the use of `print()`--as a library, these really need to either be real errors (raised as exceptions) or logged using the `warnings` module. I'll open an issue.

Issue #10, opened 2015-05-27
Merge TDSCatalog related tests


lesserwhirls:
Merge:

test_catalog.py
test_dods.py

into one test for TDSCatalog.py

PR #12, opened 2015-05-27
Remove tds subpackage and replace with catalog module.

dopplershift:
The whole library is really about talking to the TDS, so having a module about it seems redundant. Since the code that's in there is all about catalogs, move it to a `catalog.py` file.

This also merges the tests that were there, possibly addressing #10 .


dopplershift:
Must have removed some covered lines when I killed the `tds/__init__.py`.  I've now adjusted the settings so that coverage needs to go down by 0.5% before it complains.

lesserwhirls:
Looks good to me. Yes, this also addresses #10. Want to add a reference to #10 to the commit message?

dopplershift:
I'll do it in the merge.

lesserwhirls:
Ok, I did the merge...but why doesn't this show up as being your work (i.e. it;s listed as my commit, and looks like it counts towards my lines of code added/removed)?

PR #16, opened 2015-05-29
Use requests for HTTP support.

### LINKED ISSUE #8: create util for http get requests
Create a util for doing basic http get requests with the siphon user agent string using the package Requests:

http://docs.python-requests.org/en/latest/

dopplershift:
Fixes #8 .

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -1,10 +1,7 @@
-from __future__ import print_function
-from ._version import get_versions
-from .metadata import TDSCatalogMetadata
 import logging
 import xml.etree.ElementTree as ET
-
-userAgent = 'siphon (%s)' % get_versions()['version']
+from .metadata import TDSCatalogMetadata
+from .util import urlopen
 
 
 class TDSCatalog(object):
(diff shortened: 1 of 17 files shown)


dopplershift:
@lesserwhirls feel free to review, but I've found the `Session object now in the `requests` API, and I think I might want to see if that's a better fit for at least part of this (rather than `http_get`).

lesserwhirls:
Ok - why don't you check out the Session object and see how it lives up.

dopplershift:
Ok, how about now? In addition to just creating a helper to create a `Session` object with the appropriate headers (which gives all of the appropriate top-level methods, like `get`), I also removed the `basic_http_request()` method, which didn't seem to really be adding anything.

siphon/catalog.py
```diff
@@ -275,59 +272,6 @@ def __init__(self, service_node):
         self.services = services
 
 
-def basic_http_request(full_url, return_response=False):
-    r"""
```
lesserwhirls: Well that's embarrassing...what kind of doc string did I write in there! I'll tell you - the wrong one!

Issue #14, opened 2015-05-28
Only expose services enabled on a given dataset


lesserwhirls:
Currently, TDSCatalog builds a set of access urls on a dataset based on the services listed in the catalog. However, this is not quite correct. Once we are correctly parsing the metadata tags, then each dataset should only expose the services and access urls that are endabled on the dataset  (i.e. the element <serviceName>fullServices</serviceName> as listed in the dataset metadata).

lesserwhirls:
Fixed by pr-#17

Issue #6, opened 2015-05-23
Add collection of Metadata to TDSCatalog


lesserwhirls:
Add functionality to read and store metadata contained in the TDS Client Catalogs. For this, a simple dictionary, or subclass thereof, will do.

dopplershift:
Fixed in 499ae18

PR #17, opened 2015-06-11
Added metadata parsing to TDSCatalog object

lesserwhirls:
Adds the ability to parse metadata blocks from TDS Client Catalogs, and only exposes services on a dataset that are defined by the ServiceName element contained in the metadata.

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -58,7 +58,13 @@ class TDSCatalog(object):
         service_skip = 0
         for child in root.iter():
             tag_type = child.tag.split('}')[-1]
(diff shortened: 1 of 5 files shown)


lesserwhirls:
address gh-#6

lesserwhirls:
**Do not merge yet**

I want to address gh-#14 in this pull request too.

lesserwhirls:
Ok, this should be ready for review now.

dopplershift:
We probably need some examples using this. I won't hold it up for those, but that will be the best way to see how the API feels.

Issue #2, opened 2015-05-20
Add client for TDS radarServer


dopplershift:
This should be able to use `TDSCatalog` to parse the dataset.xml files and speak the proper NCSS-like query language.

dopplershift:
Fixed in #19 .

PR #19, opened 2015-06-15
Data Clients

### LINKED ISSUE #3: Add NCSS client.
This should encapsulate constructing NCSS query strings, as well as parse the `dataset.xml` file that is served up initially by the NCSS endpoint. This should also handle dumping the returned NetCDF data to a temp file and opening using `netCDF4-python`.

dopplershift:
Add support for clients to fetch data from a TDS.

Fixes #3 and #2 .


dopplershift:
Not quite done yet, but wanted to start getting some feedback. Still need:
- [ ] Radar server client
- [ ] Sphinx doc pages (docstrings done)
- [ ] Examples

dopplershift:
Not sure what the deal is with the timezone import. That works fine here, should be part of stdlib.

lesserwhirls:
What specifically in terms of metadata parsing?

dopplershift:
Making use of what you just implemented.

JohnLCaron:
"code smells" ??

lesserwhirls:
Better out than in @JohnLCaron

siphon/http_util.py
```diff
@@ -0,0 +1,476 @@
...
+    See Also
+    --------
+    http_get
+    '''
+
+    return BytesIO(create_http_session().get(url, **kwargs).content)
+
+
+def parse_iso_date(s):
+    r'''Parse a string containing an ISO-8601 formatted date
```
lesserwhirls: We may want to put this in util at some point, as I could use this to get datetime objects from the catalogs.

siphon/http_util.py
```diff
@@ -0,0 +1,476 @@
...
+        return self
+
+    def lonlat_box(self, north, south, east, west):
+        r'''Add a latitude/longitude bounding box to the query.
+
+        This adds a request for a spatial bounding box, bounded by (`south`, `north`)
+        for latitude and (`west`, `east`) for the longitude. This modifies the query
+        in-place, but returns ``self`` so that multiple queries can be chained together
+        on one line.
+
```
lesserwhirls: We should probably mention the units on latitude and longitude - I think the TDS expects degree_east and degree_north.

siphon/http_util.py
dopplershift: I say do that once we have something else to put in `util` at some point; it no longer exists right now, as all of the old functions in `util` were http-related.

siphon/http_util.py
dopplershift: Make sense.

siphon/ncss.py
```diff
@@ -0,0 +1,391 @@
...
+    # Need staticmethod to keep this from becoming a bound method, where self
+    # is passed implicitly
+    unit_handler = staticmethod(default_unit_handler)
+
+    def _get_metadata(self):
+        # Need to use .content here to avoid decode problems
+        meta_xml = self.get_path('dataset.xml').content
+        root = ET.fromstring(meta_xml)
+        self.metadata = NCSSDataset(root)
+        self.variables = set(self.metadata.variables.keys())
```
lesserwhirls: Rather than doing a set, I think we should make sure to catch any duplicate variable names that might exists. This should probably be done at the NCSSDataset level though. Of course, you can't have duplicate variables in a java NetcdfDataset, so this may be a non-issue.

dopplershift: The motivation of `set` wasn't duplicates, but sub-linear (I think it's a hash, like `dict`, so O(1)) search. If we want duplicate checking, I agree that it should be in `NCSSDataset`.

siphon/ncss.py
```diff
@@ -0,0 +1,391 @@
...
+            # go past = and ", skip final "
+            units[names[-1]] = unitstr[eq + 2:-1]
+    return names, units
+
+
+def parse_csv_dataset(data, handle_units):
+    fobj = BytesIO(data)
+    names, units = parse_csv_header(fobj.readline().decode('utf-8'))
+    arrs = np.genfromtxt(fobj, dtype=None, names=names, delimiter=',', unpack=True,
+                         converters={'date': lambda s: parse_iso_date(s.decode('utf-8'))})
```
lesserwhirls: Have you ran into the funny (i.e. NOT funny) case of a csv return from a PointFeatureCollection that is missing the "T" character in the iso time string? While that should be fixed at some point [https://bugtracking.unidata.ucar.edu/browse/TDS-670], we may want to enhance parse_iso_date to handle that.

dopplershift: Agreed, but should do with a separate issue, preferably motivated using a test.

PR #32, opened 2015-06-24
If TDSCatalog detects html, warn and try an XML link.

### LINKED ISSUE #29: Handle errors when passing HTML to `TDSCatalog`
Right now, if accidentally pass an html catalog to `TDSCatalog`, we get:
```
  File "<string>", line unknown
ParseError: mismatched tag: line 7, column 66
```
I don't know if it's a good idea to rewrite the URL, but we should at least detect it somehow (maybe using the content type on the headers) and throw a clear error.

dopplershift:
Fixes #29.

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -2,7 +2,7 @@ import logging
 import xml.etree.ElementTree as ET
 
 from .metadata import TDSCatalogMetadata
-from .http_util import urlopen
+from .http_util import create_http_session, urlopen
 
 log = logging.getLogger("siphon.catalog")
 log.setLevel(logging.WARNING)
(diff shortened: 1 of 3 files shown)


siphon/catalog.py
```diff
@@ -41,11 +41,23 @@ def __init__(self, catalog_url):
...
+
+        session = create_http_session()
+
         # get catalog.xml file
-        xml_fobj = urlopen(catalog_url)
+        resp = session.get(self.catalog_url)
+
+        # If we were given an HTML link, warn about it and try to fix to xml
+        if 'html' in resp.headers['content-type']:
+            import warnings
```
lesserwhirls: Why the warnings package and not logging.warning?

dopplershift: The logic that was suggested before was: logging for warnings of bad conditions that the user may want to know about, but can't do anything about; warnings module for warnings the user should address. 

https://docs.python.org/3/howto/logging.html

Specifically: "warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning"

dopplershift:
My bigger question, is automatically trying a new URL and setting that on the `TDSCatalog` instance a good idea?

siphon/catalog.py
lesserwhirls: Ah yes, I remember now.

lesserwhirls:
I think it's ok to try with xml if the url ends with html. It's a very common cut/paste error when trying to find a catalog url.

dopplershift:
Not just url endswith html, I replace all 'html' in the URL with 'xml'. This is important for URLs like: http://thredds.ucar.edu/thredds/catalog/grib/NCEP/GFS/Global_onedeg/catalog.html?dataset=grib/NCEP/GFS/Global_onedeg/Best

lesserwhirls:
Ah. Yes. I think that is still ok to do.

PR #33, opened 2015-06-24
RadarServer: Use metadata (Fixes #27)

### LINKED ISSUE #27: Radarserver metadata.
Now that `TDSCatalog` will parse it without raising an exception, we should parse it and pull out the variables.

dopplershift:


lesserwhirls:
Looks good - merging

Issue #20, opened 2015-06-16
Catalog API enhancement


dopplershift:
The base catalog API is a bit clunky to use; we need to add some user-friendly options. Some examples are in bird_house/threddsclient. Ideas:

- [ ] `follow` method for `CatalogRef` (see `urllib.parse.urljoin`)
- [ ] Be able to see all top level datasets
- [ ] Simple regex search for datasets (i.e. find everything with name GFS)

The driving goal is to be able to access a TDS and inquire about datasets *easily*, **without** using a web browser, only the Python API. Writing example notebooks to do some tasks, starting from the top-level `catalog.xml` is a great way to identify API clunkiness.

dopplershift:
Another thought: While all the datasets have names, I have yet to see us use the name of a dataset to pull something out of that `datasets` dictionary. That tells me that particular API is completely wrong, and at the very least that should be a list equivalent to: `list(datasets.values())`.

lesserwhirls:
PR #35 addresses "follow method for CatalogRef (see urllib.parse.urljoin)".

dopplershift:
Search might also be helpful beyond regex with something like: https://github.com/jamesturk/jellyfish

ethanrd:
Ability to search a remote catalog/search service (e.g., CSW or OpenSearch) would be great.

PR #35, opened 2015-07-02
Preserve relative urls in catalog references

ScottWales:
The leading '/' in catalog references is needed to construct the full URL path from a catalog reference.

For instance, in the catalog `http://thredds-test.unidata.ucar.edu/thredds/testDatasets.xml` the reference `TestFmrc` points to `/thredds/catalog/fmrc/ecmwf/global_2p5/catalog.xml`. Resolving the relative URL should give `http://thredds-test.unidata.ucar.edu/thredds/catalog/fmrc/ecmwf/global_2p5/catalog.xml`. Removing the `/`, as is being done in `CatalogRef.__init__()`, means the URL is constructed relative to the current catalog path, rather than being relative to the hostname.

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -145,8 +145,6 @@ class CatalogRef(object):
         """
         self.name = element_node.attrib["name"]
         self.href = element_node.attrib["{http://www.w3.org/1999/xlink}href"]
-        if self.href[0] == '/':
-            self.href = self.href[1:]
         self.title = element_node.attrib["{http://www.w3.org/1999/xlink}title"]
 
 
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -4,6 +4,12 @@ import xml.etree.ElementTree as ET
 from .metadata import TDSCatalogMetadata
 from .http_util import create_http_session, urlopen
 
+try:
+    from urlparse import urljoin
+except ImportError:
+    # Python 3
+    from urllib.parse import urljoin
+
 log = logging.getLogger("siphon.catalog")
 log.setLevel(logging.WARNING)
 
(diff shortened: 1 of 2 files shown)


dopplershift:
In principle I have no problem with this, but I see @lesserwhirls is responsible for the lines in question, so I'd like a second from him.

Thanks for the contribution and tests!

lesserwhirls:
First off, I echo @dopplershift - thank you @ScottWales for the contribution and tests!

It's a minor change, and I don't see anything wrong there. At some point soon, however, we should probably look at the module `urlparse` (python2, in 3 it is `urllib.parse`) to handle joining the fragments of the url.

We can accept this as is as far as I am concerned just to get it working now. That said - @ScottWales, are you feeling frisky, because you could try to use `urlparse` in the fix to see how it goes.

Python 2:
```
import urlparse
urlparse.urljoin(base, catalog_ref)
```

Python 3:
```
import urllib.parse
urllib.parse.urljoin(base, catalog_ref)
```

Note that `catalog_ref` can be a relative path, or a fully qualified path. If relative, it is joined to `base`; if fully qualified, it is used as is. So a bonus of using `urlparse` is that it will  "do the right thing" under the hood if the for relative and fully qualified catalogRefs.

ScottWales:
Happy to help. I've modified the pull request so that `CatalogRef.href` now stores the absolute URL to the target catalog, calculated using `urljoin` from the catalog's base URL.

I've also added the `CatalogRef.follow()` function mentioned in #20 since it has a trivial implementation, and updated the test case to use it.

lesserwhirls:
That's great @ScottWales - thank you!

PR #38, opened 2015-07-03
added Siphon + ncMWS example notebook

lesserwhirls:


dopplershift:
`%matplotlib inline` is the problem. I've avoided it in other examples. I can't remember if I have that set by default somewhere.

dopplershift:
Also might need to list owslib/cartopy as a requirement for examples somewhere.

lesserwhirls:
Should we have this as an example if there are requirements outside of those from Siphon?

lesserwhirls:
Ok, it looks like I broke the travis build by adding cartopy and oswlib to the requirements.txt file under examples. Any thoughts as to why?

dopplershift:
Build probably failed because Cartopy isn't pip installable. You could put it in a subdir and that would exclude it from test, since I do think it's useful to include.

We could, instead, look into running the examples in such a way that we could mock out such imports. Could also at that point find a way to transparently use vcrpy to eliminate remote access too.

Maybe best path is just move this for now, and open an issue to improve example testing process.

lesserwhirls:
Cool - done. New issue is #39.

PR #43, opened 2015-07-08
Created cartopy example using NCSS

jclark754:
Here's another example of using NCSS to retrieve data. This uses the NCSS class and the lonlat_box function for the query to return data in a region around Colorado and plot it to a contour map using cartopy. I made it with the intention of showing something a bit prettier at the python users workshop, but someone may find it useful. Sorry the commit below says carroty, the Github client for Mac likes to autocorrect (and apparently carroty is a word?).


dopplershift:
Replaced by #99 so that we can clean it up before putting it in.

Issue #24, opened 2015-06-17
NetCDF tempfile handling broken on windows


dopplershift:
When using the NCSS client to open data we saved to a tempfile, we get a `RuntimeError` due to "Permission denied". I'm guessing Windows doesn't like someone opening the file a second time.

We could try catching the error and opening a different way, but in that case it's a much bigger pain to ensure the file gets deleted. *sigh* Why can't windows just behave in a unix-like fashion?

This may be obviated once we get the in-memory solution in the netcdf library, but I think it will be awhile after that before it's widely available. Might punt on this until a windows user complains...

MoonRaker:
I am going to be that Windows user...

Any fixes/workarounds for this yet?

dopplershift:
Well, the workaround would be to use `get_data_raw` to get the content of the netcdf file, and save it to a local file (temp or otherwise) yourself.

However, if fixing this would be a boon to your use of siphon, I'm much more inclined to fix it know I have *actual*, not hypothetical, user(s) needing it.

wholmgren:
@dopplershift See this pvlib-python [issue](https://github.com/pvlib/pvlib-python/issues/86) for more info on how we want to use Siphon for solar power forecasts. We can also help address Siphon problems relevant to our work.

wholmgren:
also, @dopplershift, many thanks to you and Unidata for developing Siphon!

dopplershift:
Wow, that looks really cool. I'm glad you're finding Siphon so useful.  In general, if you guys just open issues for problems you find (or better yet, PRs for fixes!), that will be great.

For this issue, the big problem is that we're not windows devs on Siphon. I made windows 7 VM, but it was painful. It will definitely go faster if you can find a fix that works for you and make a PR. Otherwise, I can make a pass at a fix, but the testing loop on that will probably be much slower.

MoonRaker:
I am not sure how you want to proceed but I have made a universal fix within the siphon code that deletes the temporary file upon exit, after setting the NamedTemporaryFile argument "delete" to False. This is also possible to do with less modifications to the Siphon code by deleting the temporary file using dataset.filepath() within our own code but it still requires that the "delete" argument equal False. However, both approaches require that the dataset be closed first.

dopplershift:
The sounds like the approach I took originally (never committed), until I realized that I could re-open the file (which made the code much cleaner); it's a shame that approach doesn't work on windows.

I'll take a PR with your changes to add the exit handler

[...]

PR #56, opened 2015-09-27
Fix: ncss netcdf temp file permission error on Windows

MoonRaker:
These changes were made in response to the inability of python to open a temporary file twice on Windows systems, Permission Denied error, thus preventing ncss.get_data() queries in Siphon. [Issue #24 NetCDF tempfile handling broken on windows](https://github.com/Unidata/siphon/issues/24).

The temporary file is now persistent but is then deleted at exit. The temporary file will only be removed if .close() method is called on the netCDF Dataset. With this fix the error message disappears and all nosetests pass. It has not been tested on Linux/Unix.

Siphon v0.3.1
Code that reproduces the error (also fails 2 nosetests; test_netcdf_point, test_netcdf4_point):
best_gfs = TDSCatalog('http://thredds.ucar.edu/thredds/catalog/grib/NCEP/GFS/Global_0p5deg/catalog.xml?dataset=grib/NCEP/GFS/Global_0p5deg/Best')
best_gfs.datasets
best_ds = list(best_gfs.datasets.values())[0]
best_ds.access_urls
ncss = NCSS(best_ds.access_urls['NetcdfSubset'])
query = ncss.query()
query.lonlat_point(-110.9, 32.2).vertical_level(100000).time_range(start, end)
query.variables('Temperature_isobaric').accept('netcdf')
data = ncss.get_data(query)

...line 351, in read_netcdf
    return Dataset(tmp_file.name, 'r')
  File "netCDF4\_netCDF4.pyx", line 1616, in netCDF4._netCDF4.Dataset.__init__ (netCDF4\_netCDF4.c:9989)
RuntimeError: Permission denied

### DIFF
--- a/siphon/ncss.py
+++ b/siphon/ncss.py
@@ -1,10 +1,10 @@
 import xml.etree.ElementTree as ET
+import atexit
 from io import BytesIO
+from os import remove
+import platform
 
 import numpy as np
-import os
-import platform
-import atexit
 
 from .http_util import DataQuery, HTTPEndPoint, parse_iso_date
 from .ncss_dataset import NCSSDataset
@@ -357,17 +357,6 @@ try:
                 tmp_file.flush()
                 return Dataset(tmp_file.name, 'r')
 
-# def temp_opener(name, flag, mode=0o777):
-#   return os.open(name, flag | os.O_TEMPORARY, mode)
-
-# with tempfile.NamedTemporaryFile() as f:
-#     f.write(DATA)
-#     f.flush()
-#     with open(f.name, "rb", opener=temp_opener) as f:
-#         assert f.read() == DATA
-
-# assert not os.path.exists(f.name) 
-
 except ImportError:
     import warnings
     warnings.warn('netCDF4 module not installed. '
@@ -375,7 +364,7 @@ except ImportError:
 
 def deletetempfile(fname):
     try:
-        os.remove(fname)
+        remove(fname)
     except PermissionError:
         import warnings
         warnings.warn('temporary netcdf dataset file not deleted. '
--- a/siphon/ncss.py
+++ b/siphon/ncss.py
@@ -344,12 +344,12 @@ try:
     @response_handlers.register('application/x-netcdf')
     @response_handlers.register('application/x-netcdf4')
     def read_netcdf(data, handle_units):  # pylint:disable=unused-argument
-        osType = platform.architecture()
-        if osType[1].lower() == 'windowspe':
+        ostype = platform.architecture()
+        if ostype[1].lower() == 'windowspe':
             with NamedTemporaryFile(delete=False) as tmp_file:
                 tmp_file.write(data)
                 tmp_file.flush()
-                atexit.register(deletetempfile,tmp_file.name)
+                atexit.register(deletetempfile, tmp_file.name)
                 return Dataset(tmp_file.name, 'r')
         else:
             with NamedTemporaryFile() as tmp_file:
@@ -362,10 +362,11 @@ except ImportError:
     warnings.warn('netCDF4 module not installed. '
                   'Will be unable to handle NetCDF returns from NCSS.')
 
+
 def deletetempfile(fname):
     try:
         remove(fname)
-    except PermissionError:
+    except OSError:
         import warnings
         warnings.warn('temporary netcdf dataset file not deleted. '
                       'to delete temporary dataset file in the future '
--- a/siphon/ncss.py
+++ b/siphon/ncss.py
@@ -2,7 +2,8 @@ import xml.etree.ElementTree as ET
 from io import BytesIO
 
 import numpy as np
-from os import remove
+import os
+import platform
 import atexit
 
 from .http_util import DataQuery, HTTPEndPoint, parse_iso_date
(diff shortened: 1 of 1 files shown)


siphon/ncss.py
```diff
@@ -2,6 +2,8 @@
 from io import BytesIO
 
 import numpy as np
+from os import remove
+import atexit
```
dopplershift: For cleanliness, can you move `atexit` import to the first line, and the `from os` to just under `from io`?

siphon/ncss.py
```diff
@@ -341,15 +343,29 @@ def parse_xml_dataset(elem, handle_units):
...
     @response_handlers.register('application/x-netcdf4')
     def read_netcdf(data, handle_units):  # pylint:disable=unused-argument
-        with NamedTemporaryFile() as tmp_file:
+        with NamedTemporaryFile(delete=False) as tmp_file:
             tmp_file.write(data)
             tmp_file.flush()
+            # return Dataset(tmp_file.name, 'r') 
+
+            ## Windows work around - DGG 9.14.2015
+            atexit.register(deletetempfile,tmp_file.name)
```
dopplershift: PEP8: needs space after `,`

siphon/ncss.py
```diff
@@ -341,15 +343,29 @@ def parse_xml_dataset(elem, handle_units):
     @response_handlers.register('application/x-netcdf')
     @response_handlers.register('application/x-netcdf4')
     def read_netcdf(data, handle_units):  # pylint:disable=unused-argument
-        with NamedTemporaryFile() as tmp_file:
+        with NamedTemporaryFile(delete=False) as tmp_file:
             tmp_file.write(data)
             tmp_file.flush()
+            # return Dataset(tmp_file.name, 'r')
```
dopplershift: Looks like cruft.

siphon/ncss.py
```diff
@@ -341,15 +343,29 @@ def parse_xml_dataset(elem, handle_units):
...
-        with NamedTemporaryFile() as tmp_file:
+        with NamedTemporaryFile(delete=False) as tmp_file:
             tmp_file.write(data)
             tmp_file.flush()
+            # return Dataset(tmp_file.name, 'r') 
+
+            ## Windows work around - DGG 9.14.2015
+            atexit.register(deletetempfile,tmp_file.name)
             return Dataset(tmp_file.name, 'r')
+
```
dopplershift: Extraneous whitespace added.

siphon/ncss.py
```diff
@@ -341,15 +343,29 @@ def parse_xml_dataset(elem, handle_units):
...
     warnings.warn('netCDF4 module not installed. '
                   'Will be unable to handle NetCDF returns from NCSS.')
 
+def deletetempfile(fname):
+    try:
+        remove(fname)
+    except PermissionError:
+        import warnings
+        warnings.warn('temporary netcdf dataset file not deleted. '
+                      'to delete temporary dataset file in the future '
```
dopplershift: Could you capitalize the sentences?

dopplershift:
Awesome, thanks for the contribution. Mostly just a bunch of little (picky) stuff. This looks identical to what I had originally, before I "discovered" the cleaner approach on *nix.

What I wonder is whether we should just enable this method on windows and stick with the original method on *nix. @lesserwhirls any opinions here?

siphon/ncss.py
MoonRaker: Absolutely. What is the reasoning? Just for my future reference?

MoonRaker:
Thanks for the comments, I appreciate the input for sure. Either way is fine with me.

siphon/ncss.py
dopplershift: We group the imports (based partially on PEP8) by standard library (first full import, then partial import), then third party, then our own; arranged alphabetically. Doing it consistently makes it easy to find stuff when you have quite a few imports.

dopplershift:
I think I like the approach described at http://stackoverflow.com/a/15235559 (and mentioned in #24) better. Any thoughts? I'm willing to take a pass at implementing that approach, though it might be a week before I can do so; feel free to try it yourself if you're so inclined.

Either way, if you could test it and see if it actually works on Windows, that would be very helpful.

MoonRaker:
It seems like a cleaner approach. I'll give it a go. At first pass though the entire implementation of the temporaryfile might have to look different otherwise it will require modifying the netCDF library since that is where the second open operation takes place.

dopplershift:
*paperbag moment*

You're right, I didn't think things through. We can go with your original approach, though I'd like to leave the original code for OSX/Linux and only use the `atexit` approach on Windows.

MoonRaker:
I have updated so that the usage for OSX/Linux is the same as before and only the Windows temporary file is handled with atexit.

dopplershift:
Overall, I think we're almost there. There are some pep8 issues picked up by pylint:
```
siphon/ncss.py:347:9: N806 variable in function should be lowercase
siphon/ncss.py:352:47: E231 missing whitespace after ','
siphon/ncss.py:365:1: E302 expected 2 blank lines, found 1
```

Also need to address the remaining inline comments. The actual test failures are caused by a new version of `vcrpy`. I'll take care of fixing that, but that will mean you'll need to rebase on top of the latest master once I merge.

siphon/ncss.py
```diff
@@ -341,15 +344,33 @@ def parse_xml_dataset(elem, handle_units):
...
+
 except ImportError:
     import warnings
     warnings.warn('netCDF4 module not installed. '
                   'Will be unable to handle NetCDF returns from NCSS.')
 
+def deletetempfile(fname):
+    try:
+        remove(fname)
+    except PermissionError:
```
dopplershift: `PermissionError` it turns out is >= Python 3.3. I don't mind dropping Python 3.2, but I'm not quite prepared to drop Python 2.7. Since `PermissionError` is a subclass of `OSError`, can you use that instead?

MoonRaker: That should be fine.

dopplershift:
This looks good. Thanks for going around with me on it.

MoonRaker:
I appreciate your patience and helping me work through it.

PR #57, opened 2015-10-18
Pin vcrpy to less than 1.7

dopplershift:
Up to vcrpy 1.7.3 contains a bug that breaks return values on decorated
functions.

This was addressed in kevin1024/vcrpy#219, so we can upgrade once this sees a release.


IvanMalison:
new vcrpy release (1.7.4) fixes this

PR #86, opened 2016-01-22
Replace nosetests with py.test

### LINKED ISSUE #81: nose -> py.test
Nose is dying (https://nose.readthedocs.org/en/latest/index.html) and py.test seems to be what all the cool kids are using. Shouldn't take much to move over, but will be easier to do sooner rather than later.

dopplershift:
Fixes #81.

Nose is unmaintained and the community at large seems to be adopting py.test.

Nothing serious to convert, other than no longer using nose's helpers. py.test does sensible things with regular `assert` so this is fine.


lesserwhirls:
Cool @dopplershift - this does not look like it was a bad transition at all.

PR #98, opened 2016-04-04
Use OrderedDict for TDSCatalog

dopplershift:
This uses OrderedDict to store datasets and catalog refs. This gives those items in the order they appear in the catalog, which is more sensible.


dopplershift:
Overriding loss of coverage because it doesn't reflect results including AppVeyor, which would bring us to 95.06%.

PR #108, opened 2016-09-07
Fix for accessing THREDDS catalogs hosted at a Context Path other tha…

joelrahman:
…n /thredds

I'm using siphon to access several THREDDS instances hosted at nonstandard context paths - i.e. without /thredds/ anywhere in the URL.

The /thredds/ context is hard coded in Siphon and this causes failures in connecting to these catalogs.  Basically, siphon takes the entire catalog URL and appends the service URL to the end. (I'm taking the siphon URLs and passing them to pydap)

This is  a simple fix that works for every non-standard server I've encountered thus far and passes your tests when run on Python 3.4. 

I've worked on the assumption that the service descriptions in the catalog should always include an absolute path, which includes the context path, making it safe to take everything from the URL  except the actual path (i.e. take scheme, host, username/password, port). Let me know if you see other cases and I can take a look.

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -56,7 +56,7 @@ class TDSCatalog(object):
         """
         # top level server url
         self.catalog_url = catalog_url
-        self.base_tds_url = catalog_url.split('/thredds/')[0]
+        self.base_tds_url = _find_base_tds_url(catalog_url)
 
         session = create_http_session()
 
@@ -277,14 +277,7 @@ class Dataset(object):
                 service_name = metadata["serviceName"]
 
         access_urls = {}
-        if catalog_url.find('/thredds/') >= 0:
-            server_url = catalog_url.split('/thredds/')[0]
-        else:
-            url_components = urlparse(catalog_url)
-            if url_components.path:
-                server_url = catalog_url.split(url_components.path)[0]
-            else:
-                server_url = catalog_url
+        server_url = _find_base_tds_url(catalog_url)
 
         found_service = None
         if service_name:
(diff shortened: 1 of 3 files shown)


dopplershift:
Thanks for this. Are you aware that github is attaching these commits to a different flowmatters account rather than joelrahman?

Could you add a test validating that this works for one of the non-standard context servers? Otherwise, we risk breaking this in the future.

@lesserwhirls Any opinions on this?

dopplershift:
I can't find any other mentions of `/thredds/` in the codebase like that; and regardless of "correctness", I will say using `urlparse` seems much better than a simple string split anyway.

joelrahman:
Happy to add a test. I note that all the current catalog tests hit unidata servers. I've got no control over the THREDDS servers I'm using - they could disappear, temporarily or permanently, leaving you with intermittent tests.

Do you want the test to hit a live server, noting those risks, or should I just add a test for _find_base_tds_url?

(Won't get back to this for 24 hours)

Re the accounts - both me sorry... Seems I had my git config messed up when I cloned the fork...

dopplershift:
Re: accounts--it's up to you. The association is based on the email address, so you could change your email in your git settings and rewrite the commits (see [here](https://www.git-tower.com/learn/git/faq/change-author-name-email)). It makes no difference to me unless you want the credit or whatnot on your other account.

For tests, our tests use a library called vcrpy to record and playback http requests. If you add a new test function with a dectorator like:
```python
@recorder.use_cassette('accessing-my-thredds-server')
def test_nonstandard_server():
'Testing server on different context'
    pass
```
It will record you accessing the server the first time you run the test, and subsequent runs of the test will use the recorded data, called a casette. Just make sure you commit the cassette file, which will be called `accessing-my-thredds-server` in this case.

dopplershift:
That looks great. Thanks for the contribution and for adding the test.

PR #113, opened 2016-10-18
BUG: Make parsing more permissive.

dopplershift:
There's no reason for the client to be so pedantic. Also change some
log.error to log.warning.


dopplershift:
Overriding patch-level test check since we're just tweaking the error path here.

PR #118, opened 2017-03-05
Latest dataset issue

lesserwhirls:
In TDS 4.6, the latest service had a `name` of `Latest` and a `serviceType` of `Resolver`, at least on Unidata servers. In 5.0, the name has changed to simply `Resolver`, as part of the new Standard Service feature. The name was always configurable by the server admin, so there is no need to explicitly check the `service` elements `name` attribute.

Also in this PR is change to make invalid values for THREDDS Catalog Metadata less noisy, as they are not all that importiant for the end users to see. This second commit sets the log level to ERROR instead of WARNING.

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -400,8 +400,7 @@ def _get_latest_cat(catalog_url):
     """
     cat = TDSCatalog(catalog_url)
     for service in cat.services:
-        if (service.name.lower() == "latest" and
-                service.service_type.lower() == "resolver"):
+        if (service.service_type.lower() == "resolver"):
             latest_cat = cat.catalog_url.replace("catalog.xml", "latest.xml")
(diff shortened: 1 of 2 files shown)


lesserwhirls:
This will be a good fix to get in the next release, as the GOES-16 datasets will likely be on thredds-jumbo, which siphon cannot talk to without this fix.

lesserwhirls:
@dopplershift - no pytest? I'm guessing you ran into this at some point with MetPy?

dopplershift:
The build I just checked looks like a more serious photo if problem....I can dig in more later.

dopplershift:
This will need to be rebased on master after #119 is merged, which should fix our test problems.

dopplershift:
Alright, you should be good for a rebase now.

lesserwhirls:
:metal:

PR #124, opened 2017-03-21
Fix validation warning in tests

lesserwhirls:
We get a validation warning in our test suite due to an attribute added to a catalog that is not under our control.

### DIFF
--- a/siphon/metadata.py
+++ b/siphon/metadata.py
@@ -12,6 +12,7 @@ import logging
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.ERROR)
+log.addHandler(logging.StreamHandler())
 
 
 class _SimpleTypes(object):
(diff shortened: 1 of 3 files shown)


dopplershift:
So.....the presence of that invalid value actually served as a test of the warning code. *sigh*

dopplershift:
How would you feel about using pytest's [warning support](http://doc.pytest.org/en/latest/recwarn.html) instead?

lesserwhirls:
Hmmm...I'd rather add a new test to `test_metadata.py` that explicitly tests for the warning.

lesserwhirls:
Oh man...I think I see it now 👎

dopplershift:
Separate test is fine, but then you need to swallow the warning in the other test.

lesserwhirls:
Oops...that comment was for another issue.

Yes, a separate test that triggers and swallows the warning using pytests warning support.

dopplershift:
Can you add that here so we don't drop (meaningful) coverage?

lesserwhirls:
Yep, will do - just need to get aws going again *laughs* I'm running single threaded tonight.

lesserwhirls:
So `pytest.warns` does not seem to play nice with `logging.warning` - it wants `warnings.warn`. I seem to recall we discussed `logging` and `warnings` at some point, but I can't remember the details.

dopplershift:
So it looks like [pytest-catchlog](https://github.com/eisensheng/pytest-catchlog) is the answer (it's on PyPI and conda-forge). If you're not up for adding it, I can do it.

lesserwhirls:
I found that, too. I can add it in - I just wanted to talk to you about it before doing so. Let me update the PR.

PR #127, opened 2017-04-26
Handle more data types when parsing dataset.xml

lesserwhirls:
Before these changes, we only handled types `int` and `float` when parsing dataset.xml - everything else was left as a string and logged a warning. Even with this PR, we still do not handle:

 * Sequence
 * Structure
 * enum
 * opaque
 * object
 * char

because I do not have an example of those, and honestly have no idea what to do with some.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -60,7 +60,7 @@ sphinx_gallery_conf = {
     'examples_dirs': ['../examples'],
     'gallery_dirs': ['examples'],
     'filename_pattern': '/',
-    'mod_example_dir': 'api/generated'
+    'backreferences_dir': 'api/generated'
 }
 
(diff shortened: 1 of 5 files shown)


lesserwhirls:
Failing on docs?

dopplershift:
That should fix it, let's see...

lesserwhirls:
I did a boo boo @dopplershift. You fixed the docs and pushed a commit...I forgot you did that and used `--force` on my updates.

no-force

dopplershift:
I still had it, not that it was a hard change.

lesserwhirls:
Sorry about that...

my bad

dopplershift:
Apparently there's a new version of flake8-comprehensions out--I've pushed cleanups for the new lint.

PR #129, opened 2017-05-23
Fix build

dopplershift:
- Fix up config so flake8-quotes runs correctly
- Clean up docstrings and enable flake8-docstrings
- Move to using pytest-catchlog to capture and test log messages (this seems better supported and is recommended from pytest-dev/pytest#2079). Pytest 3.1 broke our current handling.


dopplershift:
Sphinx gallery fails due to Unicode problems. I'm waiting to see the resolution of sphinx-gallery/sphinx-gallery#246 before I decide whether to disable our windows doc builds, and hence example runs.

dopplershift:
@lesserwhirls @jrleeman One of you want to look this over? It passed before I rebased. Need this to go in before I do my next big PR with our API updates.

siphon/__init__.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) 2013-2015 University Corporation for Atmospheric Research/Unidata.
 # Distributed under the terms of the MIT License.
 # SPDX-License-Identifier: MIT
+"""Tools for accessing atmospheric and oceanic science data on remote servers."""
```
jrleeman: Isn't AOS a bit limited considering lots of folks run THREDDS? But then again this isn't a THREDDS specific library. Maybe something like : Tools for accessing remote data hosted on THREDDS and other servers? Maybe that's too general?

siphon/cdmr/coveragedataset.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) 2016 University Corporation for Atmospheric Research/Unidata.
 # Distributed under the terms of the MIT License.
 # SPDX-License-Identifier: MIT
+"""Provide access to TDS' Coverage Dataset."""
```
jrleeman: TDS' took me a couple of glances. Maybe this reads cleaner? Provide access to a THREEDS server's coverage dataset

siphon/__init__.py
lesserwhirls: The only TDSs I've encountered in the wild are AOS. I'd say it's ok for now, especially since it is kind of the target audience, and siphon is still pretty unknown (I think).

jrleeman: Fair enough

siphon/cdmr/coveragedataset.py
lesserwhirls: Yeah, it does look odd. Maybe "Provide access to **a TDS** Coverage Dataset."

siphon/catalog.py
```diff
@@ -335,15 +337,14 @@ def make_access_urls(self, catalog_url, all_services, metadata=None):
         self.access_urls = access_urls
 
     def add_access_element_info(self, access_element):
+        """Create an access method from a catalog element."""
```
lesserwhirls: Why is it that sometimes we use a regular string

~~~
"""blah blah blah"""
~~~

but others, we use a raw string (like line 347 below)

~~~
r"""blah blah blah"""
~~~

lesserwhirls:
Should have done this as a review, not single comments...sorry about that.

facepunch

siphon/catalog.py
dopplershift: At some point I was in the habit of using raw to make sure LaTeX math renders properly. Not really needed here.

siphon/cdmr/coveragedataset.py
dopplershift: Agreed. Fixed.

siphon/catalog.py
dopplershift: Nuked 'em.

lesserwhirls: I may have added some of them as well, no biggie.

dopplershift:
@jrleeman you may want to look over the last commit.

Issue #137, opened 2017-06-30
Add latest, etc. attributes to Dataset


dopplershift:
`Dataset` should have attributes for:
- `latest`
- `best`
- `collection` (for full collection on Grib or point feature collection)

These should raise a useful error message for datasets that don't have such virtual datasets.

dopplershift:
Latest was done in #140. Leaving this open for best and full collection.

Issue #136, opened 2017-06-30
Simplify access to download services


dopplershift:
No more having people directly use `access_urls`.
![img_0243](https://user-images.githubusercontent.com/221526/27754402-a61fa75e-5da7-11e7-8215-85bce5b58313.jpg)

jrleeman:
Spelling of methods:
`remote_subset`
`remote_access`

Issue #135, opened 2017-06-30
Add datetime filtering of datasets


jrleeman:
We need to add methods to access datasets with datetime objects:

* `filter_time_nearest` returns a single dataset with the time closest to that requested
* `filter_time_range` returns a list of datasets falling inclusively into the time range. Start time is the first positional argument. The end time is keyword and defaults to `utcnow`.

Use a series of regular expressions to parse the filename string. 

These leave us open to future expansion for where we want to source the times from (file name, time dimension, etc.)

Issue #134, opened 2017-06-30
Improved dataset indexing


dopplershift:
Replace the `TDSCatalog.datasets` dictionary with a thin data structure that allows indexing:
- positionally
- by dataset name

```python
cat = TDSCatalog(...)
cat.datasets[0]
cat.dataset['GEFS_Global_1p0deg_Ensemble_ana_20170617_0000.grib2']
```

This can be done in a backwards compatible manner by having the new `datasets` attribute (of type `DatasetCollection`?) inherit `dict`.

PR #140, opened 2017-07-20
Improve API

dopplershift:
A lot of changes to smooth over the API:
- Add index-based access to collections of datasets and catalog refs (#134)
- Add date time-based filtering of those collections (#135)
- Add methods to simplify remote access (#136)
- Add a `latest` attribute to simplify access to latest dataset (#137)

This leaves adding the `format` option to `subset` to future work. I also punted on best/collection since those aren't easily programmatically determined.

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -24,6 +24,18 @@ log.addHandler(logging.StreamHandler())  # Python 2.7 needs a handler set
 log.setLevel(logging.ERROR)
 
 
+class IndexableMapping(OrderedDict):
+    """Extend ``OrderedDict`` to allow index-based access to values."""
+
+    def __getitem__(self, item):
+        """Return an item either by index or name."""
+        try:
+            item + ''  # Raises if item not a string
+            return super(IndexableMapping, self).__getitem__(item)
+        except TypeError:
+            return list(self.values())[item]
+
+
 class TDSCatalog(object):
     """
     Parse information from a THREDDS Client Catalog.
@@ -79,9 +91,9 @@ class TDSCatalog(object):
         root = ET.fromstring(resp.text)
         self.catalog_name = root.attrib.get('name', 'No name found')
 
-        self.datasets = OrderedDict()
+        self.datasets = IndexableMapping()
         self.services = []
-        self.catalog_refs = OrderedDict()
+        self.catalog_refs = IndexableMapping()
         self.metadata = {}
         self.ds_with_access_elements_to_process = []
         service_skip_count = 0
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -448,6 +448,120 @@ class Dataset(object):
         url_path = access_element.attrib['urlPath']
         self.access_element_info[service_name] = url_path
 
+    def download(self, filename):
+        """Download the dataset to a local file.
+
+        Parameters
+        ----------
+        filename : str
+            The full path to which the dataset will be saved
+
+        """
+        with self.remote_open() as infile:
+            with open(filename, 'wb') as outfile:
+                outfile.write(infile.read())
+
+    def remote_open(self):
+        """Open the remote dataset for random access.
+
+        Get a file-like object for reading from the remote dataset, providing random access,
+        similar to a local file.
+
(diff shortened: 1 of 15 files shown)


siphon/catalog.py
```diff
@@ -342,6 +453,120 @@ def add_access_element_info(self, access_element):
...
+        """
+        if service == 'CDMRemote':
+            from .cdmr import Dataset as CDMRDataset
+            provider = CDMRDataset
+        elif service == 'OPENDAP':
+            try:
+                from netCDF4 import Dataset as NC4Dataset
+                provider = NC4Dataset
+            except ImportError:
+                raise ValueError('OPENDAP access requires netCDF4-python to be installed.')
```
jrleeman: Shouldn't this be using `ModuleNotFoundError`?

siphon/catalog.py
```diff
@@ -342,6 +453,120 @@ def add_access_element_info(self, access_element):
...
+        elif service == 'NetcdfSubset':
+            from .ncss import NCSS
+            provider = NCSS
+        elif service == 'HTTPServer':
+            provider = urlopen
+        else:
+            raise ValueError(service + ' does not have a built-in handler')
```
jrleeman: Maybe a bit different spelling? `not a supported access method in Siphon`?

siphon/catalog.py
```diff
@@ -24,6 +26,106 @@
 log.setLevel(logging.ERROR)
 
 
+class IndexableMapping(OrderedDict):
+    """Extend ``OrderedDict`` to allow index-based access to values."""
+    def __getitem__(self, item):
+        """Return an item either by index or name."""
+        try:
+            item + ''  # Raises if item not a string
```
jrleeman: Why not check the type? It's a bit more look before you leap I suppose.

dopplershift: That's pretty much it. And I don't like having to figure out what the right type is (though here it's probably `basestring`).

siphon/catalog.py
dopplershift: I consider it a `ValueError` for the specified service, since the user asked for one they can't use.

siphon/catalog.py
dopplershift: Definitely better. Have gone with: `is not an access method supported by Siphon`.

dopplershift:
Line length is set to 95 in the configs already. Where am I missing the problem about line length?

siphon/catalog.py
jrleeman: Fair enough

siphon/catalog.py
jrleeman: Not sure I agree on that still. They CAN use the service (i.e. it's available on the server), but they need a module installed locally.

jrleeman:
Codacy is loudly complaining about the line length. Can it be silenced?

Looks like a few minor flake issues are hanging around.

siphon/catalog.py
dopplershift: Alright, you convinced me.

dopplershift: So....`ModuleNotFoundError` seems to be Python 3 only. `ImportError`?

dopplershift:
Found the codacy one, it uses prospector, which is in `.prospector.yml`.

jrleeman:
Looks like we may have to go with `ImportError`... Thanks 2.7!

jrleeman:
Note: Safe to ignore codacy - does not approve of Ryan's ask-forgiveness string type check.

dopplershift:
And QuantifiedCode wants me to document an internal method.

dopplershift:
Coverage just showed me that I was doing cdmremote wrong...oops. MOAR TESTS!

lesserwhirls:
Do you guys generally do reviews before or after the automated tests are passing?

dopplershift:
Some of both? In general it's fine to do before, since review is more about catching things the computers don't catch, especially for us who do this a lot.

dopplershift:
Not to mention, as backed up as AppVeyor is on our account right now, it's likely I can address any comments now before this PR runs again...

PR #143, opened 2017-07-21
BUG: Fix empty set returned from validate_query (Fixes #102)

### LINKED ISSUE #102: NCSS validate_query with no variables
``` python
from siphon.ncss import NCSS
ncss = NCSS('http://thredds.ucar.edu/thredds/ncss/nws/metar/ncdecoded/Metar_Station_Data_fc.cdmr')
query = ncss.query()
ncss.validate_query(query)
```

returns `set()`

Looks like the nice short-circuit behavior results in the return of the empty variable query--I think a call to `bool()` is in order.

dopplershift:
Manually cast set of variables to bool when returning.

### DIFF
--- a/siphon/ncss.py
+++ b/siphon/ncss.py
@@ -88,7 +88,7 @@ class NCSS(HTTPEndPoint):
 
         """
         # Make sure all variables are in the dataset
-        return query.var and all(var in self.variables for var in query.var)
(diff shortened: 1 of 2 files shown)


dopplershift:
Travis failure is due to failing example from down radar, and is "addressed" in #140 . I'll merge if everything else passes.

Issue #130, opened 2017-05-31
Port upper air support from MetPy


dopplershift:
Remote server access should be the purview of Siphon, not MetPy, so we should move the upper air code (that reads from Wyoming and Iowa St.) here.

jrleeman:
We need to make sure this is bumped to the next release per the MetPy depreciation warning.

dopplershift:
Punting from 0.5 to next release so we can resolve:
1. Need to either port metpy's CDM layer or re-write using netCDF4-python in-memory support. I lean to the latter, but at that point we should probably just rely upon netCDF4.
1. Need to figure out the package layout for this, especially with an eye towards adding additional services like these, such as AWS/S3-based REST services

dopplershift:
Closed by #147.

PR #147, opened 2017-08-23
Move upper air data capability into Siphon

jrleeman:
Addresses #130 and moves upper air data support into Siphon. Still a sketch given that we have some questions on where exactly this should live. I've used the in memory capability of netCDF4-python here, but without wrapping the `Variable` like we do in MetPy, units are a no go. Then again, we'd be depending on MetPy's unit registry at that point, so maybe we return a plain netCDF `Dataset`? I'm thinking a tool in MetPy that goes through and deals with appending units from metadata might be useful anyway.


siphon/upperair.py
```diff
@@ -0,0 +1,282 @@
...
+        Defaults to 'wyoming'.
+
+    kwargs
+        Arbitrary keyword arguments to use to initialize source
+
+    Returns
+    -------
+        :class:`metpy.io.cdm.Dataset` containing the data
+
+    .. deprecated:: 0.6.0
```
lesserwhirls: Do we need this warning mirrored in siphon?

lesserwhirls:
Do we have tests in MetPy that we can move over to Siphon?

siphon/upperair.py
jrleeman: Nope, we can get rid of it.

jrleeman:
There are some tests that will need to be re-mocked. Especially my test on server return codes.

siphon/simplewebservice/wyoming.py
```diff
@@ -0,0 +1,149 @@
+# Copyright (c) 2013-2015 University Corporation for Atmospheric Research/Unidata.
+# Distributed under the terms of the MIT License.
+# SPDX-License-Identifier: MIT
+"""Read upper air data from the Wyoming archives."""
+
+from ..http_util import HTTPEndPoint
+from io import BytesIO
+import numpy as np
+import pandas as pd
+from metpy.calc import get_wind_components
```
jrleeman: Planning on porting `get_wind_components` from metpy into an `_tools`.

siphon/simplewebservice/wyoming.py
```diff
@@ -0,0 +1,149 @@
...
+
+    kwargs
+        Arbitrary keyword arguments to use to initialize source
+
+    Returns
+    -------
+        :class:`pandas.DataFrame` containing the data
+
+    """
+    endpoint = WyomingUpperAir('http://weather.uwyo.edu/')
```
jrleeman: I don't like the URL here, I'm feeling like the object setup isn't ideal here.

siphon/simplewebservice/wyoming.py
```diff
@@ -0,0 +1,149 @@
...
+        return data_dict
+
+    data = dict()
+    for key, name in [('p', 'pressure'),
+                      ('z', 'height'),
+                      ('t', 'temperature'),
+                      ('td', 'dewpoint')]:
+        key_data, key_units = info[key]
+
+        insert_with_units(data, name, key_data, key_units)
```
jrleeman: Need to add u, v, speed, direction still.

siphon/simplewebservice/wyoming.py
```diff
@@ -0,0 +1,149 @@
...
+
+    return pd.DataFrame(data=data)
+
+
+class WyomingUpperAir(HTTPEndPoint):
+    """Download and parse data from the University of Wyoming's upper air archive."""
+
```
dopplershift: I think the way to go is to add a `__init__` that takes no arguments and passes the url to:
```python
super(WyomingUpperAir, self).__init__('myurlhere')
```
Should also include `'cgi-bin/sounding'` in that url.

siphon/simplewebservice/wyoming.py
```diff
@@ -0,0 +1,149 @@
...
+            Site id for which data should be downloaded
+        region : str
+            The region in which the station resides. Defaults to `naconf`.
+
+        Returns
+        -------
+        a file-like object from which to read the data
+
+        """
+        path = ('cgi-bin/sounding?region={region}&TYPE=TEXT%3ALIST'
```
dopplershift: Let's encompass this in a `Query` object for Wyoming. Might need to refactor `DataQuery` to allow a query object for this to inherit useful behavior.

siphon/upperair.py
```diff
@@ -0,0 +1,279 @@
+# Copyright (c) 2013-2015 University Corporation for Atmospheric Research/Unidata.
```
dopplershift: I assume this file is going away.

siphon/simplewebservice/wyoming.py
```diff
@@ -0,0 +1,149 @@
...
+                '&YEAR={time:%Y}&MONTH={time:%m}&FROM={time:%d%H}&TO={time:%d%H}'
+                '&STNM={stid}').format(region=region, time=time, stid=site_id)
+
+        resp = self.get_path(path)
```
dopplershift: With everything above, this becomes `self.get_query(query)`.

siphon/simplewebservice/wyoming.py
```diff
@@ -0,0 +1,149 @@
...
+                'for station {stid}.'.format(time=time, region=region,
+                                             stid=site_id))
+
+        # Since the archive text format is embedded in HTML, look for the <PRE> tags
+        data_start = resp.text.find('<PRE>')
+        data_end = resp.text.find('</PRE>', data_start)
+
+        # Grab the stuff *between* the <PRE> tags -- 6 below is len('<PRE>\n')
+        buf = resp.text[data_start + 6:data_end]
+        return BytesIO(buf.encode(encoding='UTF-8'))
```
dopplershift: Can probably just move the contents of `parse` here since we're not trying to have the same common structure as we did in MetPy. Might be able to leverage pandas' own parsing code to do this work too.

siphon/simplewebservice/wyoming.py
dopplershift: Put it in the constructor for the class...see below.

siphon/simplewebservice/wyoming.py
```diff
@@ -0,0 +1,149 @@
...
+    Returns
+    -------
+        :class:`pandas.DataFrame` containing the data
+
+    """
+    endpoint = WyomingUpperAir('http://weather.uwyo.edu/')
+    fobj = endpoint.get_data(time, site_id)
+    info = endpoint.parse(fobj)
+
+    direction, spd, spd_units = info['wind']
```
dopplershift: I think all of the pandas manipulation should be in `get_data` in the WyomingUpperAir class.

siphon/simplewebservice/wyoming.py
```diff
@@ -0,0 +1,149 @@
...
+"""Read upper air data from the Wyoming archives."""
+
+from ..http_util import HTTPEndPoint
+from io import BytesIO
+import numpy as np
+import pandas as pd
+from metpy.calc import get_wind_components
+
+
+def get_upper_air_data(time, site_id):
```
dopplershift: I think:
1. This function should be a `staticmethod` on `WyomingUpperAir`
2. It should just wrap the process of building a query and calling `get_data`.

dopplershift:
Happy to discuss more...probably easier to get into details now that you have a concrete picture.

jrleeman:
Will need to have a discussion about using query tomorrow, but I have reworked the parsing completely and it's now pretty robust as far as I can tell.

jrleeman:
Ready for re-review @dopplershift

jrleeman:
Also notice that some of the indexes in the test changed from MetPy - we're returning the full data frame, even if the first rows (lowest altitudes) are all NaN's. I think that's fine, but am up for disagreement.

dopplershift:
Is there an easy way with pandas to remove empty rows?

jrleeman:
Technically not an empty row as the populate height still.... I'll see if there is an easy way to cut if all other fields are NaN

jrleeman:
See what you think about this. If we like it, I'll fixup. Not the nicest solution, but the end result is nice. You must drop the row only when all the columns specified are NaN, then reindex so the index column starts a zero and is continuous.

jrleeman:
Seems verbose for a simple task. i.e. manually specifying the index rest and needed to give a kwarg to prevent maintaining the old index as a column. It is a one-liner, just seemed overly long for a simple task.

dopplershift:
Me thinks someone's standards are a tad high. 😈  Lest I remind you this is how we did this before:
```python
            if any(np.invert(np.isnan(values[1:]))):
                arr_data.append((level,) + values)
```
Not exactly the bastion of simple code either.

jrleeman:
Don't burst my idealistic bubble 😄 - just fixed up, pending CI, we're good to go.

dopplershift:
I would have figured that bubble had been burst many times over the last 6 months. 😁

PR #152, opened 2017-09-06
BUG: Fix access_urls for some services

dopplershift:
Catalogs are allowed to specify the name of a service that is contained
within a compound service. Siphon was not properly creating the access
urls dictionary for this case.

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -410,15 +410,18 @@ class Dataset(object):
             The top level server url
         all_services : List[SimpleService]
             list of :class:`SimpleService` objects associated with the dataset
-        metadata : TDSCatalogMetadata
+        metadata : dict
             Metadata from the :class:`TDSCatalog`
 
         """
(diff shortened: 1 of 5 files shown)


siphon/catalog.py
```diff
@@ -410,15 +410,18 @@ def make_access_urls(self, catalog_url, all_services, metadata=None):
...
 
         """
-        all_service_dict = {service.name: service for service in all_services}
-        service_name = None
-        if metadata:
-            if 'serviceName' in metadata:
-                service_name = metadata['serviceName']
+        all_service_dict = {}
+        for service in all_services:
+            all_service_dict[service.name] = service
```
jrleeman: Should this be in an else? (i.e. should the compound service be in the dictionary or just it's sub services?)

dopplershift: You want both. I initially had it as an `else`, and it (rightly) failed tests.

jrleeman: 👍

PR #154, opened 2017-09-14
Printing/repr improvements

dopplershift:
This has a few improvements to string representations and printing to help when working interactively (i.e. in the notebook)

- Add `__repr__` (which is just the same as `__str__`) to `Dataset` and friends for the CDM Remote interface. This just makes it easier to get to them in the notebook
- Add a `__str__` method to `catalog.datasets` to make it easier to print out what datasets are available without needing to manually call `list`.


dopplershift:
Well that makes me feel good, a combination of deleting our old Python 3.7 netCDF4 package and fixing a stupid test gets us green on the Python 3.7 nightly build.

PR #162, opened 2017-10-19
Doc updates

dopplershift:
- Add a DOI and citation information
- Update some resources, including linking to the gallery
- Hide the table of contents from the landing page but leave it in the side bar
- Add the contributor's guide to the web documentation

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,0 +1,1 @@
+../CONTRIBUTING.md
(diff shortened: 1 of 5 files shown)


dopplershift:
You mean like we already do under "Related Projects"? 😈

jrleeman:
Fine 🙄

dopplershift:
So we've been linking to a non-existent list of authors in our contributor's guide. I fixed that, but accidentally pushed that directly to master.

Rebased this on top so that hopefully it will go green.

I swear, there's something broken in anything that's not automatically checked--in this case a dead link in the contributor guide.

PR #163, opened 2017-10-24
MNT: Fix style lint from flake8 3.5

dopplershift:
It now warns about variables named 'l', since they are easily confused
with '1'. Changing all of these did seem to result in better code
regardless.


dopplershift:
Yay, green!

PR #177, opened 2018-01-09
Adds acis.py for ACIS Web Services functionality

WEP11:
Adds acis.py, which provides a function for connecting to ACIS Web Services. This function is Python 3 compatible, but will need to use urllib2 to become Python2.7 compatible.


siphon/acis.py
```diff
@@ -0,0 +1,51 @@
...
+
+    req = urllib.request.Request(base_url+method,
+                          params, {'Content-Type': 'application/json'})
+    if method == 'MultiStnData':
+        timeout=300
+    else:
+        timeout=60
+
+    try:
+        response = urllib.request.urlopen(req, timeout=timeout)  # 10-Minute Timeout
```
dopplershift: I don't think that comment is correct any more 😁 

Would also need to be updated to use `requests`.

siphon/acis.py
```diff
@@ -0,0 +1,51 @@
...
+        return json.loads(response.read())
+    except urllib.request.HTTPError as error:
+        if error.code == 400:
+            print(error.msg)
```
dopplershift: Instead of catching the error, printing the message, and returning `False`, I think it's better to just let the error bubble to the top. That turns the code for users from (currently):
```python
response = acisRequest(...)
if response == False:
    # Handle error
else:
    # Do stuff with data
```
into
```python
try:
    response = acisRequest(...)
    # Do stuff with data
except HTTPError:
    # Handle errors
```
It's shorter code in the library, just as easy for users, and doesn't involve having changing return types. If you're worried about having multiple exception types, you could create a custom `Exception` subclass and raise that from inside the function.

siphon/acis.py
```diff
@@ -0,0 +1,51 @@
+import json
+import urllib.request
+import urllib.parse
+import json
+import socket
+
+def acisRequest(method, params):
```
dopplershift: How about `acis_request`? That matches our existing naming structure.

siphon/acis.py
```diff
@@ -0,0 +1,51 @@
...
+    will have an error text in the return data produced by the remote API.
+
+    Parameters
+    ----------
+    method - The Web Services request method (StnMeta, StnData, MultiStnData, etc)
+    params - A dictionary of parameters (see API documentation)
+
+    Returns
+    ----------
+    Connection Success: A dictionary that is derived from JSON data from the remote API
```
dopplershift: Would need to be updated based on discussion below. Also Sphinx will pitch a fit unless the length of "---------" matches the length of "Returns".

siphon/acis.py
```diff
@@ -0,0 +1,51 @@
...
+    based on the given method (StnMeta,StnData,MultiStnData,GridData,General)
+    and parameters string. Information about the parameters can be obtained at:
+    http://www.rcc-acis.org/docs_webservices.html
+
+    If a connection to the API fails, then it returns False. A failed request
+    will have an error text in the return data produced by the remote API.
+
+    Parameters
+    ----------
+    method - The Web Services request method (StnMeta, StnData, MultiStnData, etc)
```
dopplershift: Parameters docs look like:
```
method : str
    The Web Services request method (StnMeta, StnData, MultiStnData, etc.)
```

siphon/acis.py
```diff
@@ -0,0 +1,51 @@
...
+    ----------
+    Connection Success: A dictionary that is derived from JSON data from the remote API
+    Connection Failure: False
+
+    """
+    params = json.dumps(params).encode('utf8')
+
+    base_url = 'http://data.rcc-acis.org/'  # ACIS Web API URL
+
+    req = urllib.request.Request(base_url+method,
```
dopplershift: Do you think you could adapt this to use the `requests` module? We're already using it internally, it would solve your Python 2 problem, and it sets our own "User-Agent" header. I'm happy to look into how to do this if you can clarify how the `data` argument gets used.

siphon/acis.py
WEP11: That certainly can be done.

siphon/acis.py
WEP11: The data argument is simply a JSON object containing parameters (Station ID, dates, variables, etc) for the HTTP request. I'm more than happy to adapt it, there was no reason for using urllib other than that's what was used in the old Python 2 version of this code.

WEP11:
I am working on some use case examples. Will those go into `/examples`?

siphon/acis.py
dopplershift: Ok, then I'll leave you to it, but feel free to scream if you run into problems.

dopplershift:
Yeah, I think that makes sense, though if you have several, you could do `examples/acis`.

WEP11:
Latest push adds two examples to the documentation and finalizes some of the acis_request function code. Exception handling is certainly open to feedback.

siphon/simplewebservice/acis.py
```diff
@@ -0,0 +1,65 @@
...
+        raise ACIS_API_Exception("Connection Timeout")
+    except requests.exceptions.TooManyRedirects:
+        raise ACIS_API_Exception("Bad URL. Check your ACIS connection method string.")
+    except ValueError:
+        raise ACIS_API_Exception("No data returned! The ACIS parameter dictionary\
+                                may be incorrectly formatted")
+
+class ACIS_API_Exception(Exception):
+    """
+
```
dopplershift: Can you kill the blank line?

siphon/simplewebservice/acis.py
```diff
@@ -0,0 +1,65 @@
...
+    base_url = 'http://data.rcc-acis.org/'  # ACIS Web API URL
+
+    if method == 'MultiStnData':
+        timeout=300
+    else:
+        timeout=60
+
+    try:
+        response = requests.post(base_url+method, json=params, timeout=timeout)
```
dopplershift: Can you add spaces around the `+`?

dopplershift: Also, if you change `requests.post()` to `create_http_session().post()` (after doing `from ..http_util import create_http_session`), you'll get e.g. "Siphon(0.6.0)" in the logs for ACIS. Not sure if you'd be interested in seeing how much use the service gets from us--or if you even have access to those logs yourself. This one really doesn't matter to me since we'll never see the logs.

examples/acis/Mapping_Example.py
```diff
@@ -0,0 +1,147 @@
...
+# data.
+#
+# Now let's look at how our data has changed. Rather than having a just a 'meta'
+# and 'data' component, we have a new one called 'smry'. As you've guessed,
+# this contains our summary information (also in the order we requested it).
+# By specifying 'smry_only', there is no 'data' component. If we also wanted
+# all 4 days of data, we would simply remove that parameter.
+#
+# To wrap up this example, we will finally plot our precipitation sums and
+# departures onto a map using CartoPy and MetPy. To do this we will utilize
```
dopplershift: It doesn't look like you're actually using MetPy here.

siphon/simplewebservice/acis.py
```diff
@@ -0,0 +1,65 @@
...
+        return response.json()
+    except requests.exceptions.Timeout:
+        raise ACIS_API_Exception("Connection Timeout")
+    except requests.exceptions.TooManyRedirects:
+        raise ACIS_API_Exception("Bad URL. Check your ACIS connection method string.")
+    except ValueError:
+        raise ACIS_API_Exception("No data returned! The ACIS parameter dictionary\
```
dopplershift: Single quotes, please. Also, instead of the line continuation (`\`), you can just end the line with a `'`, and then open a new string with a `'` on the next line; since it's within the `()`, Python will concatenate them for you. So:
```python
raise ACIS_API_Exception('No data returned! The ACIS parameter dictionary'
                         'may be incorrectly formatted')
```

dopplershift:
We'll also need to add a test. Probably easiest to start with one from `siphon/tests/test_wyoming.py`. The `recorder.use_cassette` the first time you run the test will record all of the web requests and responses; future runs of the test (when it can find the recorded data--so it needs to be included in a commit) will use the recorded data so that we don't have tests that need a proper network, server availability, etc. You'll need to make sure vcrpy is installed.

If setting that up is a problem, I'm happy to jump in and help out.

examples/acis/Mapping_Example.py
WEP11: You caught me! I was, but then I noticed the implementation I had was poor. Looks like I forgot to proofread that part of the tutorial afterwards.

siphon/simplewebservice/acis.py
WEP11: Neat! I'll have to double check with the developer that manages the servers to make sure, but I know it'd be interesting to see. (We're more than happy to share those statistics if we get them)

WEP11:
Alright, I have a unit test file that runs fine when run I run the individual function. How can I run the test to generate the recorded data you mentioned? Thanks!

WEP11:
This version adds the unit test, fixes the formatting errors mentioned, and adds the unit test data to the fixtures folder. I completely missed that the data was already present in my fork (I imagined a gitignore for some reason). I'll monitor the bots to see what they find, thanks!

siphon/tests/test_acis.py
```diff
@@ -0,0 +1,41 @@
+'''Test ACIS Web Services API Access'''
+from siphon.simplewebservice.acis import acis_request
+from siphon.testing import get_recorder
+
+recorder = get_recorder(__file__)
+
+@recorder.use_cassette('acis_request')
+def test_acis():
```
dopplershift: Can you break this up into 4 individual tests, one for each request method?

WEP11: I certainly can

siphon/tests/test_acis.py
```diff
@@ -0,0 +1,54 @@
...
+    assert data['data'][0]['data'][0] == '37.5'
+    assert data['data'][1]['meta']['uid'] == 13284
+    assert data['data'][1]['data'][0] == '49.0'
+
+
+def test_acis_griddata():
+    """Testing ACIS Gridded Data request."""
+    data = acis_request('GridData', {'loc': '-95.36, 29.76', 'sdate': '2000-01',
+                        'edate': '2000-07', 'grid': '3', 'elems': [
+                        {'name': 'maxt', 'interval': 'mly', 'reduce': 'max',
```
WEP11: I'm unsure of how to take care of this PEP8 error. I thought line 44 might have been too long, but instead I created the E128 error.
`siphon/tests/test_acis.py:44:25: E122 continuation line missing indentation or outdented
siphon/tests/test_acis.py:45:25: E128 continuation line under-indented for visual indent`

siphon/tests/test_acis.py
```diff
@@ -0,0 +1,54 @@
+"""Test ACIS Web Services API Access."""
+from siphon.simplewebservice.acis import acis_request
+from siphon.testing import get_recorder
+
+recorder = get_recorder(__file__)
+
+
+@recorder.use_cassette('acis_request')
```
WEP11: I used one "cassette" for all 5 tests.. Is that alright? They're all JSON strings.

dopplershift: I think that's fine for now. If we ever need to regenerate, we'll have to separate them, since the cassette is only ever generated from a single test.

siphon/tests/test_acis.py
dopplershift: Try this:
```python
    data = acis_request('GridData',
                        {'loc': '-95.36, 29.76', 'sdate': '2000-01', 'edate': '2000-07',
                         'grid': '3', 'elems': [{'name': 'maxt', 'interval': 'mly',
                                                 'reduce': 'max', 'smry': 'max'}]})
```
Arguments to the function, keys in the dictionary, items in the list--they line up when wrapped.

examples/acis/Mapping_Example.py
```diff
@@ -0,0 +1,149 @@
...
+# their areas!
+
+lat = []
+lon = []
+pcpn = []
+pcpn_dep = []
+
+for stn in myData['data']:
+    # Skip threaded stations! They have no lat/lons
+    if stn['meta']['sids'][-1][-1] == '9':
```
dopplershift: Maybe
```
if stn[‘meta’][‘sids’][-1].endswith(‘9’)
```

examples/acis/Basic_Overview.py
```diff
@@ -0,0 +1,95 @@
...
+# a string, so you should convert those back into numbers before use.
+#
+# *Note: Missing data is recorded as M!
+
+stnName = myData['meta']['name']
+
+avgt = []
+dates = []
+for obs in myData['data']:
+    if obs[0][-2:] == '01':
```
dopplershift: How about:
```python
If obs[0].endswith(‘01’)
```
?

examples/acis/Mapping_Example.py
```diff
@@ -0,0 +1,149 @@
...
+        continue
+    # Skip stations with missing data
+    if stn['smry'][0] == 'M' or stn['smry'][1] == 'M':
```
dopplershift: What comes before “M” if it’s at index 1? White space? If so, what about:
```
If stn[‘smry’].lstrip().startswith(‘M’)
```
?

WEP11: In this case 'smry' is actually a list of summary values. So in this case the 'smry' would have one value for the total precip and another for the departure from normal.

WEP11:
That push should take care of most of the changes you found.. I 'dismissed' your review I guess?

dopplershift:
I have it set to clear the review when new code is uploaded.

This is good. Thanks! 🎉 Congrats on your PR!

Issue #179, opened 2018-01-25
Failure to get access urls


jrleeman:
In moving things from NOMADS to NCEI we're failing to return any access urls. Not sure if it's a siphon or NCEI issue yet.

Broken example:

```python
from siphon.catalog import TDSCatalog
base_url = 'https://www.ncei.noaa.gov/thredds/catalog/namanl/'
dt = datetime(2011, 4, 27)
cat = TDSCatalog('{}{dt:%Y%m}/{dt:%Y%m%d}/catalog.xml'.format(base_url, dt=dt))
ds = cat.datasets[0]
print(ds.access_urls)
```

dopplershift:
So the problem is caused by the following:
```xml
<metadata inherited="true">
<serviceName>ALL</serviceName>
<authority>gov.noaa.ncdc</authority>
<dataType>GRID</dataType>
<documentation type="funding">
U.S. Climate Program Office (reanalysis grant), EMC, CPC, NCEP Central Operations (NCO), and NESDIS
</documentation>
<documentation type="processing_level">Level 4</documentation>
<documentation type="rights">
Electronic downloads of the data are free, however fees apply for data certification and distribution of the data on physical media. Fees vary based on order specifications.
</documentation>
<creator>
<name vocabulary="GCMD">DOC/NOAA/NWS/NCEP</name>
<contact url="http://www.ncep.noaa.gov/" email="http://www.ncep.noaa.gov/mail_liaison/"/>
</creator>
<keyword vocabulary="GCMD">
GOVERNMENT AGENCIES-U.S. FEDERAL AGENCIES, DOC, NOAA, DOC/NOAA/NESDIS/NCEI, National Centers for Environmental Information, NESDIS, U.S. Department of Commerce, EARTH SCIENCE SERVICES, MODELS, EARTH SCIENCE REANALYSES/ASSIMILATION MODELS
</keyword>
<metadata xlink:href="http://gis.ncdc.noaa.gov/geoportal/rest/document?id={6439CC43-0208-4AD6-BF6F-48F586F7541D}" xlink:title="ISO 19115-2:2009(E) - Collection Level Metadata"/>
<project>North American Mesoscale</project>
<publisher>
<name vocabulary="GCMD">DOC/NOAA/NESDIS/NCEI</name>
<contact url="https://www.ncei.noaa.gov/" email="ncei.info@noaa.gov"/>
</publisher>
</metadata>
```
Note the nested metadata element. Where this breaks is:
[...]

PR #188, opened 2018-02-24
BUG: Fix embedded metadata elements

lesserwhirls:
Addresses the case where a metadata element pointing to an external metadata document, via xlink, is a child of a parent metadata element.

Fixes Unidata/siphon#179

### DIFF
--- a/siphon/metadata.py
+++ b/siphon/metadata.py
@@ -11,6 +11,9 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.ERROR)
 log.addHandler(logging.StreamHandler())
 
+xlink_href_attr = '{http://www.w3.org/1999/xlink}href'
+xlink_title_attr = '{http://www.w3.org/1999/xlink}title'
+
 
 class _SimpleTypes(object):
     def __init__(self):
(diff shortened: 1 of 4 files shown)


siphon/metadata.py
```diff
@@ -495,6 +500,13 @@ def _get_tag_name(element):
             element_name = element.tag
         return element_name
 
+    @staticmethod
+    def _is_external_metadata_doc(element):
+        attributes = element.attrib
+        has_xlink_title = xlink_title_attr in attributes
+        has_xlink_href = xlink_href_attr in attributes
+        return (has_xlink_title and has_xlink_href)
```
dopplershift: No need for parens.

siphon/tests/test_catalog.py
```diff
@@ -302,3 +302,13 @@ def test_catalog_ref_str():
...
+
+
+@recorder.use_cassette('ncei_embedded_metadata')
+def test_catalog_with_embedded_metadata_elements():
+    """Test catalog with embedded metadata elements."""
+    url = 'https://www.ncei.noaa.gov/thredds/catalog/namanl/201802/20180220/catalog.xml'
+    cat = TDSCatalog(url)
+    md = cat.metadata
+    assert 'external_metadata' in md
+    assert 'serviceName' in md
```
dopplershift: This is sufficient to prove that it's working?

dopplershift: ok, works for me.

PR #191, opened 2018-03-09
Modify state borders for Cartopy 0.16

jrleeman:
Looks like we only have one example using state borders, but this should be the last thing we maintain that needs these updates.


jrleeman:
Still needs lots of test cleanup, but I wanted to run this against CI to see if any weirdness from legacy or others had crept in.

PR #196, opened 2018-03-20
BUG: Not all ACIS tests were using VCR

dopplershift:
Fix by adding the missing decorator.


dopplershift:
I caught it because we had a freak test failure on Python 3.4 due to this. Once again...how did this pass our code review?

PR #199, opened 2018-03-27
UW sounding regions

### LINKED ISSUE #176: Add `region` kwarg to documentation for upper air data requests
It's useful, but not usefully documented, only showing up as `*kwargs`

jrleeman:
Closes #176 and improves the region functionality. Users can specify human readable names instead of the abbreviations (which aren't listed anywhere I've found). The region will also be case insensitive. Adds a test for making sure the region is valid or a helpful `KeyError` is raised.

### DIFF
--- a/siphon/simplewebservice/wyoming.py
+++ b/siphon/simplewebservice/wyoming.py
@@ -46,10 +46,10 @@ class WyomingUpperAir(HTTPEndPoint):
 
         """
         endpoint = cls()
-        df = endpoint._get_data(time, site_id, **kwargs)
+        df = endpoint._get_data(time, site_id)
         return df
 
-    def _get_data(self, time, site_id, region='naconf'):
+    def _get_data(self, time, site_id):
         r"""Download and parse upper air observations from an online archive.
 
         Parameters
@@ -61,15 +61,12 @@ class WyomingUpperAir(HTTPEndPoint):
             The three letter ICAO identifier of the station for which data should be
             downloaded.
 
-        region
-            Region to request data from
-
         Returns
         -------
             :class:`pandas.DataFrame` containing the data
 
         """
-        raw_data = self._get_data_raw(time, site_id, region)
+        raw_data = self._get_data_raw(time, site_id)
         soup = BeautifulSoup(raw_data, 'html.parser')
         tabular_data = StringIO(soup.find_all('pre')[0].contents[0])
         col_names = ['pressure', 'height', 'temperature', 'dewpoint', 'direction', 'speed']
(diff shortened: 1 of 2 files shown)


siphon/simplewebservice/wyoming.py
```diff
@@ -99,13 +108,31 @@ def _get_data_raw(self, time, site_id, region='naconf'):
...
+        regions = {'north america': 'naconf', 'south america': 'samer',
+                   'south pacific': 'pac', 'new zealand': 'nz',
+                   'antarctica': 'ant', 'arctic': 'np', 'europe': 'europe',
+                   'africa': 'africa', 'southeast Asia': 'seasia',
+                   'mideast': 'mideast'}
+
+        try:
+            region = regions[region.lower()]
+        except KeyError:
+            raise(KeyError('Invalid region {}. Valid regions are '
```
dopplershift: `raise` is not a function. Also, the indentation should line up with the `'` in `KeyError('`.

siphon/simplewebservice/wyoming.py
```diff
@@ -36,6 +36,12 @@ def request_data(cls, time, site_id, **kwargs):
             The three letter ICAO identifier of the station for which data should be
             downloaded.
 
+        region (optional) : str
+            Region for the sounding data. Valid regions are `North America`,
+            `South America`, `South Pacific`, `New Zealand`, `Antarctica`,
+            `Arctic`, `Europe`, `Africa`, `Southeast Asia`, `Mideast` Defaults
```
dopplershift: Period after `Mideast`. Also, should probably replace backticks with single quotes...I think.

jrleeman:
Interestingly enough, no matter what the region is, you can get data for any station anywhere. I believe this is a fluke in the web API as I verified the behavior in the browser. Selecting a region changes the map, but doesn't appear to actually limit the station search area in the database. We should probably ping Larry on this.

siphon/simplewebservice/wyoming.py
```diff
@@ -99,13 +108,31 @@ def _get_data_raw(self, time, site_id, region='naconf'):
...
+        regions = {'north america': 'naconf', 'south america': 'samer',
+                   'south pacific': 'pac', 'new zealand': 'nz',
+                   'antarctica': 'ant', 'arctic': 'np', 'europe': 'europe',
+                   'africa': 'africa', 'southeast Asia': 'seasia',
+                   'mideast': 'mideast'}
+
+        try:
+            region = regions[region.lower()]
+        except KeyError:
+            raise KeyError("Invalid region {}. Valid regions are 'North America'"
```
dopplershift: should probably raise a `ValueError` instead of `KeyError`, since the error is that the user passed in a bad value.

jrleeman:
Larry says they do not use region except for the map on the website, so we don't need the ability to pass it. I've just defaulted to naconf and removed the ability to specify a region.

PR #198, opened 2018-03-27
Update LICENSE and general maintenance

jrleeman:
This now does a few things:
- Closes #175 by changing us to BSD 3 Clause.
- Updates `AUTHORS.txt`
- Fixes missing copyright notices in ACIS files
- Fixes doc errors by running `doc8` before building docs, so that we don't check generated RST
- Fix versioned docs by adding missing configuration in `conf.py`

PR #200, opened 2018-03-28
Doc enhancements

### LINKED ISSUE #181: Add source button to documentation
In MetPy I turned on links to source from the API docs. We should mirror that here.

jrleeman:
General documentation enhancements
* Closes #181 by adding source link in docs.
* Add transparent and white background Siphon logos in 150x150.
* Use logos on doc homepage and as example thumbnails in the gallery.


docs/conf.py
```diff
@@ -82,7 +85,7 @@
 
 # General information about the project.
 project = 'Siphon'
-copyright = '2015, Unidata'
+copyright = '2018, Unidata'
```
lesserwhirls: Following Unidata/siphon#198, should this be changed from "Unidata" to "Siphon Contributors"?

dopplershift: It should be, but I did all that in the last PR...

jrleeman: Yep - this was mostly done yesterday afternoon before you did all that. I'll kill that commit.

PR #209, opened 2018-03-31
Cleanup doc config

dopplershift:
Pull in stuff from metpy and make it match metpy better:

- "Improve this page" link (thanks @leouieda)
- Add siphon logo to upper left corner

We need to figure out how to standardize this config, templates, etc.


dopplershift:
Done. Also updated main page text a bit to remove part about us being focused on Unidata tech.

PR #210, opened 2018-04-02
Add attaching units to upper air example

### LINKED ISSUE #202: Add example of pulling out data with unit from soundings
This is a block of boiler plate that is a pain to write regularly. Let's at least put it in the docs.

jrleeman:
Closes #202

### DIFF
--- a/examples/upperair/Wyoming_Request.py
+++ b/examples/upperair/Wyoming_Request.py
@@ -11,6 +11,8 @@ the Wyoming upper air archive.
 
 from datetime import datetime
 
+from metpy.units import units
+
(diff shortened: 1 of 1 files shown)


examples/upperair/Wyoming_Request.py
```diff
@@ -37,3 +37,13 @@
 
 ####################################################
 print(df.units['pressure'])
+
+####################################################
+# Units can then be attached to the values from the dataframe.
+from metpy.units import units
```
dopplershift: Let's make the bot happy and put this at the top. The alternative is `# noqa`.

PR #216, opened 2018-04-09
BUG: Fix NCEI NCSS access

### LINKED ISSUE #215: Suport netCDFServer name
When using NCEI data, they use the older name for NCSS. This makes workshop materials more complicated and requires too much knowledge of the data source for users. We thought this was in 0.7 😭

dopplershift:
Fixes #215. Adds support for NetcdfServer, when available, to be used for NCSS.

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -355,6 +355,8 @@ class Dataset(object):
 
     """
 
+    ncssServiceNames = ('NetcdfSubset', 'NetcdfServer')
+
     def __init__(self, element_node, catalog_url=''):
         """Initialize the Dataset object.
 
@@ -540,7 +542,9 @@ class Dataset(object):
         Parameters
         ----------
         service : str, optional
-            The name of the service for subsetting the dataset. Defaults to 'NetcdfSubset'.
+            The name of the service for subsetting the dataset. Defaults to 'NetcdfSubset'
+            or 'NetcdfServer', in that order, depending on the services listed in the
+            catalog.
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


siphon/catalog.py
```diff
@@ -581,7 +584,7 @@ def access_with_service(self, service):
                 provider = NC4Dataset
             except ImportError:
                 raise ImportError('OPENDAP access requires netCDF4-python to be installed.')
-        elif service == 'NetcdfSubset':
+        elif service in ('NetcdfSubset', 'NetcdfServer'):
```
lesserwhirls: Would it be useful to make `('NetcdfSubset', 'NetcdfServer')` a variable (`ncssServiceNames`?), in case we need to do this special snowflake check elsewhere?

dopplershift: Sure. I've updated the PR to do that and use it exclusively.

lesserwhirls:
Do we want to add a test to keep codecov happy, or just merge it in?

dopplershift:
I've added a test, but I'm also checking something. It looks like we have some lines in the tests that are not executed; in the tests for exceptions it looks like we have the checks on the messages within the `raises` blocks, which causes the checks to not be executed.

dopplershift:
Ok, now the checks are at least finding the deficiencies.

dopplershift:
This should get it all happy again. @lesserwhirls can you check the change I made to the geospatial coverage metadata tag test? It wasn't ever actually finding anything, so no checks were made. I've modified it to check things, but it seems...unconventional. Also, that test class has members that are not actually used, so you may want to give the whole test a once-over.

lesserwhirls:
Well...those tests are...umm...who wrote that?

yikes

I think you're change looks ok, but I think we need to revisit the `TDSCatalogMetadata` class big time. This will take a tikihut day, for sure. I'm guessing no one is using `TDSCatalogMetadata`, because WTF did I do there.

dopplershift:
Just less beer this time until *after* you finish the code. 😁

lesserwhirls:
Sad that is, I don't think there was beer involved with that code!

no-beer-crazy

dopplershift:
Just remember: I gave you an excuse. :grin:

lesserwhirls:
...and I apprecaite that, but this garbage is all mine. I mean seriously, how is this useful???

~~~python
entry['zpositive']['zpositive']
~~~

wut

dopplershift:
Ready for someone to give this the green thumbs up. Leave it open, though, so we can use it for an asana demo.

PR #223, opened 2018-04-17
Fix error in IGRA2-derived relative humidity columns

DanielWatkins:
Adjusted the relative_humidity column converters so instead of importing an int as in the raw file, it converts to a decimal like it's supposed to.


dopplershift:
pip 10 doesn't seem to like our CI setup...lemme see what's up.

dopplershift:
If you merge/rebase on master, this should run cleanly now.

dopplershift:
Thanks for fixing this up!

PR #231, opened 2018-05-23
Maintenance

dopplershift:
Trying to get the build passing again:
- Fix new lint from pep8-naming
- Drop Python 3.4
- Install sphinx-gallery from git master to get around incompatibility with sphinx 1.7 and fix doc build


dopplershift:
Ping @lesserwhirls so that we can merge and get things back to green.

Issue #224, opened 2018-04-19
Add default filename for download


dopplershift:
Right now, `Dataset.download()` requires a filename to be passed in. We should make that optional, defaulting to the value of `Dataset.name`. So download should look roughly like
```python
    def download(self, filename=None):
        if filename is None:
            filename = self.name
        with self.remote_open() as infile:
            with open(filename, 'wb') as outfile:
                outfile.write(infile.read())
```
with of course an updated docstring.

PR #232, opened 2018-05-29
make filename parameter optional in catalog.download

haileyajohnson:
Solves #224

### DIFF
--- a/siphon/tests/test_catalog_access.py
+++ b/siphon/tests/test_catalog_access.py
@@ -83,6 +83,22 @@ def test_dataset_download(nids_url):
 
 
 @recorder.use_cassette('cat_to_open')
+def test_dataset_default_download(nids_url):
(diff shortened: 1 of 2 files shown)


siphon/tests/test_catalog_access.py
```diff
@@ -82,6 +82,22 @@ def test_dataset_download(nids_url):
         os.remove(temp)
 
 
+@recorder.use_cassette('cat_to_open')
+def test_dataset_default_download(nids_url):
```
dopplershift: Only wondering if we should avoid managing the temporary directory manually and instead use the pytest [`tmpdir`](https://docs.pytest.org/en/latest/tmpdir.html) fixture.

jrleeman:
Merging as-is as the fixture isn’t easily applied in this particular test.

PR #233, opened 2018-06-19
include <dataset> ID in siphon.Dataset

juozasg:
The 'ID' attribute from THREDDS catalog is useful for some purposes and should be parsed

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -369,6 +369,10 @@ class Dataset(object):
 
         """
         self.name = element_node.attrib['name']
+        if 'ID' in element_node.attrib:
+            self.id = element_node.attrib['ID']
+        else:
+            self.id = self.name
         if 'urlPath' in element_node.attrib:
             self.url_path = element_node.attrib['urlPath']
         else:


lesserwhirls:
Good catch! My only comment would be that since `ID` is an optional attribute on the `dataset` element in THREDDS client catalog [spec](https://www.unidata.ucar.edu/software/thredds/current/tds/catalog/InvCatalogSpec.html#dataset), I think we should leave it unset if it does not exist, rather than using the value of  the `name` attribute.

siphon/catalog.py
```diff
@@ -369,6 +369,10 @@ def __init__(self, element_node, catalog_url=''):
 
         """
         self.name = element_node.attrib['name']
+        if 'ID' in element_node.attrib:
+            self.id = element_node.attrib['ID']
+        else:
+            self.id = self.name
```
jrleeman: I tend to agree with @lesserwhirls here. I think setting to `None` is probably the best default behavior.

jrleeman:
@juozasg - Would you mind signing the CLA? We can fixup everything else.

jrleeman:
Closed in favor of #251

PR #234, opened 2018-06-29
Implement all-station IEM Upper Air request

jthielen:
This PR adds in the ability to request upper air data for all stations for a given time from the Iowa State archive via the `request_all_data` method. Additionally, a separate commit in the PR adds the simple web services other than `wyoming` to the docs.


siphon/simplewebservice/iastate.py
```diff
@@ -107,7 +138,7 @@ def _get_data(self, time, site_id):
                     'time': None}
         return df
 
-    def _get_data_raw(self, time, site_id):
+    def _get_data_raw(self, time, site_id, pressure):
```
jrleeman: What if `pressure` defaulted to None?

PR #236, opened 2018-07-09
treat services as case-insensitive

haileyajohnson:
Use case-insensitive comparisons for service names to handle catalog xmls which define either "OPenDAP", "OPENDAP", or "opendap", for example.


dopplershift:
Would it be better to make `dict` and `str` subclasses that are case-insensitive and use those to store members on `TDSCatalog` and friends?

haileyajohnson:
Closed because it was easier to start from scratch with @dopplershift 's suggestion.

PR #237, opened 2018-07-10
add case-insensitive strings and dicts for service types

haileyajohnson:
Use case-insensitive comparisons for service names to handle catalog xmls which define either "OPenDAP", "OPENDAP", or "opendap", for example.

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -135,12 +135,12 @@ class DatasetCollection(IndexableMapping):
     __repr__ = __str__
 
 
-def _try_lower(obj):
+def _try_lower(arg):
     try:
-        obj = obj.lower()
+        arg = arg.lower()
     except (TypeError, AttributeError, ValueError):
         pass
-    return obj
+    return arg
 
 
 class CaseInsensitiveStr(str):
@@ -179,6 +179,14 @@ class CaseInsensitiveStr(str):
         return str.__ne__(self._lowered, _try_lower(other))
 
 
+def _try_case_insensitive(arg):
+    try:
+        arg = CaseInsensitiveStr(arg)
+    except (TypeError, AttributeError, ValueError):
+        pass
+    return arg
+
+
 class CaseInsensitiveDict(dict):
     """Extend ``dict`` to use a case-insensitive key set."""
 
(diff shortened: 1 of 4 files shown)


siphon/catalog.py
```diff
@@ -640,7 +729,7 @@ def __init__(self, service_node):
 
     def is_resolver(self):
         """Return whether the service is a resolver service."""
-        return self.service_type == 'Resolver'
+        return CaseInsensitiveStr(self.service_type) == 'Resolver'
```
dopplershift: If you replace line 726 above with:
```python
self.service_type = CaseInsensitiveStr(service_node.attrib['serviceType'])
```
you could avoid doing the conversion here and at line 546 above.

siphon/catalog.py
```diff
@@ -447,15 +535,15 @@ def make_access_urls(self, catalog_url, all_services, metadata=None):
...
+        access_urls = CaseInsensitiveDict({})
         server_url = _find_base_tds_url(catalog_url)
 
         # process access urls for datasets that reference top
         # level catalog services (individual or compound service
         # types).
         if service_name in all_service_dict:
             service = all_service_dict[service_name]
-            if service.service_type != 'Resolver':
+            if CaseInsensitiveStr(service.service_type) != CaseInsensitiveStr('Resolver'):
```
dopplershift: Shouldn't you be able to leave `'Resolver'` alone?

lesserwhirls:
@haileyajohnson - an example for testing log messages can be found here:

https://github.com/Unidata/siphon/blob/20c017a65d1c98ae33f5c91fca04dbe0fd60f447/siphon/tests/test_metadata.py#L546-L551

We talked about using `DEBUG` messages for cases like the `pass` in `siphon/catalog.py` (in your PR):

~~~python
def _try_case_insensitive(arg):
    try:
        arg = CaseInsensitiveStr(arg)
    except (TypeError, AttributeError, ValueError):
        pass
~~~

Currently, the logging level in the catalog test script is set to `WARNING`, so we'd need to change that to `DEBUG`:

https://github.com/Unidata/siphon/blob/20c017a65d1c98ae33f5c91fca04dbe0fd60f447/siphon/tests/test_catalog.py#L15

Although I think you can set the logging level inside the test by calling using `caplog.set_level`, like shown [here](https://docs.pytest.org/en/latest/logging.html#caplog-fixture).

All of that said, maybe it's not even worth adding a logging message there and just go with the silent pass? Yeah, it drops code coverage by a fraction, but then again, it's not too hard to add something? I don't know - what say you, @dopplershift and @jrleeman ?

siphon/tests/test_catalog_access.py
```diff
@@ -141,3 +141,40 @@ def test_dataset_no_handler(nids_url):
...
+                'text/tropical/atlantic/hdob/High_density_obs_20170824.txt')
+    assert ds.access_urls == {'HTTPSERVER': http_url}  # test __eq__
+    assert ds.access_urls[wrong_case_key] == http_url  # test __getitem___
+    assert wrong_case_key in ds.access_urls  # test __contains__
+    ds.access_urls[wrong_case_key] = test_string
+    assert ds.access_urls[wrong_case_key] == test_string  # test __setitem__
+    try:
+        del ds.access_urls[wrong_case_key]  # test__delitem
+    except KeyError:
+        pytest.fail('KeyError: ' + wrong_case_key)
```
haileyajohnson: I suspect this is not the best way to test delete, anyone know of a better way?

lesserwhirls:
Looks good to me!

PR #239, opened 2018-07-17
Add capability to deal with stations with no identifier from Wyoming

### LINKED ISSUE #238: Accessing Wyoming Soundings no ICAO Station ID
So I was attempting to access data from March of 1976 from the University of Wyoming Sounding archive and when going back that far, not every station is associated with a three-letter ICAO ID on the site. A number of the stations are listed by number. Okay not terrible, but not great. I put the number in instead of the station ID and it finds the right data file, but due to the lack of a Station ID, the parsing info from below "Station information and sounding indices" it doesn't work correctly (http://weather.uwyo.edu/cgi-bin/sounding?region=naconf&TYPE=TEXT%3ALIST&YEAR=1976&MONTH=03&FROM=0400&TO=0400&STNM=72349) compared to (http://weather.uwyo.edu/cgi-bin/sounding?region=naconf&TYPE=TEXT%3ALIST&YEAR=1976&MONTH=03&FROM=0400&TO=0400&STNM=72456) the different is in the lack of the "Station Identifier:" line in the 72349 station file.

There might be a relatively easy way to make the parsing of that first line more amenable to working with a set of data that doesn't contain a Station Identifier line. The following code will give the error that results from the previous description.

```python
from datetime import datetime

from siphon.simplewebservice.wyoming import WyomingUpperAir

date = datetime(1976, 3, 4, 0)

data = WyomingUpperAir.request_data(date, '72349')
```

jrleeman:
Closes #238 

Deals with missing lines from the wyoming archive when the station has a number, but no identifier.

### DIFF
--- a/siphon/simplewebservice/wyoming.py
+++ b/siphon/simplewebservice/wyoming.py
@@ -81,6 +81,12 @@ class WyomingUpperAir(HTTPEndPoint):
         # Parse metadata
         meta_data = soup.find_all('pre')[1].contents[0]
         lines = meta_data.splitlines()
+
+        # If the station doesn't have a name identified we need to insert a
+        # record showing this for parsing to proceed.
+        if 'Station number' in lines[1]:
+            lines.insert(1, 'Station identifier: ')
+
         station = lines[1].split(':')[1].strip()
         station_number = int(lines[2].split(':')[1].strip())
         sounding_time = datetime.strptime(lines[3].split(':')[1].strip(), '%y%m%d/%H%M')
--- a/siphon/tests/fixtures/wyoming_sounding_no_station
+++ b/siphon/tests/fixtures/wyoming_sounding_no_station
@@ -1,0 +1,110 @@
+interactions:
+- request:
+    body: null
+    headers:
+      Accept: ['*/*']
+      Accept-Encoding: ['gzip, deflate']
+      Connection: [keep-alive]
+      User-Agent: [Siphon (0.7.0+34.gb41b402.dirty)]
+    method: GET
+    uri: http://weather.uwyo.edu/cgi-bin/sounding/?region=naconf&TYPE=TEXT%3ALIST&YEAR=1976&MONTH=03&FROM=0400&TO=0400&STNM=72349
+  response:
+    body: {string: "<HTML>\n<TITLE>University of Wyoming - Radiosonde Data</TITLE>\n<LINK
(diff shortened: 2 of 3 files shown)


siphon/simplewebservice/wyoming.py
```diff
@@ -81,6 +81,12 @@ def _get_data(self, time, site_id):
         # Parse metadata
         meta_data = soup.find_all('pre')[1].contents[0]
         lines = meta_data.splitlines()
+
+        # If there are only 26 lines, the station doesn't have a name identified
+        # and we need to insert a record showing this for parsing to proceed.
+        if len(lines) == 26:
```
dopplershift: You couldn't instead check:
```python
lines[1].startswith('Station identifier')
```
?

jrleeman: Could sure

jrleeman:
Happy to change it, though I don't really see it being more or less brittle as any other weirdness in the footer will still break things. I'll revise now.

siphon/simplewebservice/wyoming.py
dopplershift: Just seems like it could change from 26 lines for a variety of reasons.

jrleeman: Sure, but we'll break on all of  them 😭 
Push coming.

jrleeman:
Ok, not quite as easy as your `startswith` because we have a lot of leading whitespace and the field is totally missing, not just unpopulated. Anywho, this should do the trick.

dopplershift:
Right, if it's missing then (now accounting for whitespace) `lines[1].strip().startswith('Station identifier:')` should return `False`, no? Your new way is fine, I just wasn't sure if there was something else I'm missing.

jrleeman:
The whitespace... It would have to have 28 leading whitespaces on the string. Ugly.

dopplershift:
That’s fine. But my original suggestion was to look for “Station Identifier” to be missing from `lines[1]`, yours is to look only for “Station number” there. I just was confused as to why my approach, which is checking directly for the case that you’re solving, wouldn’t work.

Either way is fine, and I don’t want to bike she’d this to death. Just trying to see if I’m missing something.

jrleeman:
I just think this is a bit clearer and they both accomplish the same task with the same complexity as far as I can tell. Tomato vs Tomato?

PR #241, opened 2018-08-01
Add NDBC simple web service

jrleeman:
This simple web service provides access to the latest observations file for all buoys, as well as the realtime buoy data for 9 different data types.


siphon/simplewebservice/ndbc.py
```diff
@@ -0,0 +1,560 @@
...
+
+class NDBC(HTTPEndPoint):
+    """Download and parse data from the National Data Buoy Center."""
+
+    def __init__(self):
+        """Set up endpoint."""
+        super(NDBC, self).__init__('https://www.ndbc.noaa.gov/')
+
+    @classmethod
+    def station_observations(cls):
```
dopplershift: Um, this method is empty.

siphon/simplewebservice/ndbc.py
```diff
@@ -0,0 +1,560 @@
...
+                      'rain': 'hourly rain data'}
+        available_data = {}
+        buoy_url = 'https://www.ndbc.noaa.gov/data/realtime2/' + buoy + '.'
+        for key in file_types:
+            if endpoint._check_if_url_valid(buoy_url + key):
+                available_data[key] = file_types[key]
+        return available_data
+
+    @classmethod
+    def raw_buoy_data(cls, buoy, type='txt'):  # noqa: A002
```
dopplershift: How about going with `data_type` instead of `type` so we can avoid overriding a built-in (and the flake8-plugin checking for it).

siphon/simplewebservice/ndbc.py
jrleeman: Left over from spelling internal debating 😄

jrleeman:
I've addressed the changes, switched to using pandas table parser vs fixed width, added several examples, and tied up a few loose ends here.

examples/ndbc/buoy_met_request.py
```diff
@@ -0,0 +1,45 @@
...
+
+####################################################
+# Get a pandas data frame of all of the observations, meteorological data is the default
+# observation set to query.
+df = NDBC.realtime_observations('41002')
+df.head()
+
+####################################################
+# Let's make a simple time series plot to checkout what the data look like.
+fig = plt.figure(figsize=(12, 10))
```
dopplershift: I think I prefer this:
```python
fig, ax = plt.subplots(3, 1, figsize=(12, 10))
ax2b = ax[2].twinx()
ax[1].plot(...)
...
```

What do you think?

siphon/simplewebservice/ndbc.py
```diff
@@ -0,0 +1,556 @@
...
+                     'wind_direction': 'degrees',
+                     'wind_speed': 'meters/second',
+                     'wind_gust': 'meters/second',
+                     'pressure': 'hPa',
+                     'air_temperature': 'degC',
+                     'water_temperature': 'degC',
+                     '3hr_pressure_tendency': 'hPa',
+                     'time': None}
+
+        df = pd.read_fwf(StringIO(content), skiprows=2,
```
dopplershift: Intentional that this is still using `fwf`?

dopplershift:
What did the table parser get you?

jrleeman:
Table parser avoided the weirdness of the wind speed cutting off the tens digit unless there is 10 m/s or greater in the first 100 rows.... because the file is justified strangely.

siphon/simplewebservice/ndbc.py
jrleeman: nope

examples/ndbc/buoy_met_request.py
jrleeman: I don't love that personally - it's CS pure, but not how a scientist would implement a plot. I often named my axes how they would be labeled in the figure (a, b, etc)

dopplershift: If you want to label them that way, great. I have a pretty intense aversion to using numbers in variable names where an array will do—especially when you can get an array out of an API call. This is literally an example from my Pythonic plotting workshop unit.

You can still accomplish what you want with:
```python
fig, (a, b, c) = plt.subplots(3, 1, figsize=(12, 10))
```
How’s that read to you?

I’m not interested in how a “typical” scientist would do this—I’m interested in implementing best practice.

jrleeman: Then we'll need to change how it's done elsewhere in our workshop materials :)

PR #243, opened 2018-08-13
ENH: Add the option to return xarray from catalog helpers (Fixes #229)

### LINKED ISSUE #229: Add xarray option for `remote_access()`
Now that we're trying to use more xarray, it would be *really* nice to return xarray from remote access. This should be trivial since there are ways to get xarray to handle both CDMRemote and OPeNDAP.

dopplershift:
For OPeNDAP and CDMRemote.

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -595,7 +595,7 @@ class Dataset(object):
         """
         return self.access_with_service('HTTPServer')
 
-    def remote_access(self, service=None):
+    def remote_access(self, service=None, use_xarray=None):
         """Access the remote dataset.
 
         Open the remote dataset and get a netCDF4-compatible `Dataset` object providing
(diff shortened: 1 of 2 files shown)


siphon/catalog.py
```diff
@@ -670,14 +670,29 @@ def access_with_service(self, service):
...
-                from netCDF4 import Dataset as NC4Dataset
-                provider = NC4Dataset
-            except ImportError:
-                raise ImportError('OPENDAP access requires netCDF4-python to be installed.')
+            if use_xarray:
+                try:
+                    import xarray as xr
+                    provider = xr.open_dataset
+                except ImportError:
+                    raise ImportError('OPENDAP access needs xarray to be installed.')
```
jrleeman: Should this be OPENDAP needs xarray or something specific to using the xarray accessor like 'xarray must be installed to use xarray access method.' ?

dopplershift: Sure. Fixed.

PR #251, opened 2018-10-15
Set dataset ID attribute

jrleeman:
Supersedes #233

### DIFF
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -644,8 +644,8 @@ class Dataset(object):
             else:
                 raise RuntimeError('Subset access is not available for this dataset.')
         elif service not in self.ncssServiceNames:
-            raise ValueError(service + ' is not a valid service for subset. Options are: ' +
-                             ', '.join(self.ncssServiceNames))
+            raise ValueError(service + ' is not a valid service for subset. Options are: '
+                             + ', '.join(self.ncssServiceNames))
 
         return self.access_with_service(service)
 
--- a/siphon/catalog.py
+++ b/siphon/catalog.py
@@ -454,10 +454,8 @@ class Dataset(object):
 
         """
         self.name = element_node.attrib['name']
-        if 'urlPath' in element_node.attrib:
-            self.url_path = element_node.attrib['urlPath']
-        else:
-            self.url_path = None
+        self.id = element_node.attrib.get('ID', None)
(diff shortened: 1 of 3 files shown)


siphon/catalog.py
```diff
@@ -454,6 +454,12 @@ def __init__(self, element_node, catalog_url=''):
 
         """
         self.name = element_node.attrib['name']
+
+        if 'ID' in element_node.attrib:
+            self.id = element_node.attrib['ID']
```
dopplershift: `self.id = element_node.attrib.get('ID', None)` would be shorter... 😁

dopplershift: Could technically do that to the `'urlPath'` block below if you're feeling motivated...

siphon/catalog.py
```diff
@@ -454,8 +454,10 @@ def __init__(self, element_node, catalog_url=''):
 
         """
         self.name = element_node.attrib['name']
+
         if 'urlPath' in element_node.attrib:
             self.url_path = element_node.attrib['urlPath']
+            self.id = element_node.attrib.get('ID', None)
```
dopplershift: Did you really mean to move this line *into* the block? What I meant was that you could do, outside any if block
```python
self.id = element_node.attrib.get('ID', None)
self.url_path = element_node.attrib.get('urlPath', None)
```

siphon/tests/test_catalog.py
```diff
@@ -268,9 +268,10 @@ def test_simple_service_within_compound():
     url = ('http://thredds-test.unidata.ucar.edu/thredds/catalog/noaaport/text/'
            'tropical/atlantic/hdob/catalog.xml')
     cat = TDSCatalog(url)
-    assert (cat.datasets[0].access_urls ==
-            {'HTTPServer': 'http://thredds-test.unidata.ucar.edu/thredds/fileServer/noaaport/'
-                           'text/tropical/atlantic/hdob/High_density_obs_20170824.txt'})
+    assert (cat.datasets[0].access_urls
+            == {'HTTPServer': 'http://thredds-test.unidata.ucar.edu/thredds/'
+                              'fileServer/noaaport/text/tropical/atlantic/hdob/''
```
dopplershift: Oops, note the `''` at the end of the line.

jrleeman:
Ok - I think everything is happy here now @dopplershift

PR #256, opened 2018-11-06
MNT: Fix bad escape codes

dopplershift:
Need to use raw strings now for strings with things like `\s`, even though it isn't a valid escape code.


dopplershift:
ping @jrleeman

Issue #259, opened 2018-11-20
Support SPC reports


jrleeman:
Would be nice to easily download SPC storm reports. I've done this nearly trivially with pandas, but the automatic URL generation from a datetime would be nice. Problems do occasionally arise with poorly formatted data - I think from commas in the report text (a CSV file itself). Should probably run a year's worth of reports to make sure it all works as planned.

jthielen:
IEM has [some endpoints for LSRs](http://mesonet.agron.iastate.edu/json/#Local+Storm+Reports+for+a+given+period). Would those be useful for this, or should this just be directly accessing the CSV files from SPC?

jrleeman:
Functionality for the LSRs would be interesting, but aren't those distinct from SPC?  (I'm unsure here)

jthielen:
I'm not sure either. @akrherz Would you know?

akrherz:
I believe SPC's dataset has some QC done to the raw LSRs, could try pinging @pmarshwx here to see how much difference there.

akrherz:
thanks @pmarshwx for chiming in.  I do have a question tho as the official, finalized reports are Storm Data.  Are you going to somehow translate the Storm Data into LSR format?

pmarshwx:
We would not recreate the LSR format. Rather we would update the CSV files to the official record. 

We are still in the early stages of figuring out the best way to proceed.

akrherz:
sorry @pmarshwx , I used poor wording.  What I meant was that there isn't a one-to-one mapping of Storm Data attributes to those found in the LSR reports.  Some of the CSV columns would have to be missing, right?

jrleeman:
@pmarshwx - also I've run onto a few cases where report text had commas in it which screwed up parsing, just FYI

pmarshwx:
@akrherz: We are not 100% sure what the end will look like. It's possible we just use a subset of new making an exact replica, or just make sure that the old is a subset of the new. It's all still in the talk phase.

@jrleeman: Those comma's cause me all sorts of problems. I have some code that starts with pandas and falls back to the native CSV module I could share.

jrleeman:
That would be great @pmarshwx - here or via email would be fine.

Issue #257, opened 2018-11-12
Sounding Error Message


jrleeman:
Using the IAState upper air request the error message when data are not present has a bug and truncates the station name:
```
“ValueError: No data available for 2011-04-14 12Z for station OU.”
```

zbruick:
This will be closed by #267, so I'm adding this to the 0.9 milestone.

dopplershift:
Confirmed fixed by #267.

Issue #228, opened 2018-05-18
NHC Archive Reader


jrleeman:
Moved from https://github.com/Unidata/MetPy/issues/22

Pull out I/O code from: https://github.com/Unidata/HurricaneTracker

dopplershift:
Another option is NCEI's [IBTrACS](https://www.ncdc.noaa.gov/ibtracs/index.php), which is "Providing tropical cyclone best track data in a centralized location to aid our understanding of the distribution, frequency, and intensity of tropical cyclones worldwide."

PR #267, opened 2019-06-19
Spc

AodhanSweeney:
Adding SPC


lesserwhirls:
I had no idea Siphon was failing in so many places. It's hard for me to tell for sure, but it seems like it's not been passing for some time now. The good news is that your two classes are not causing any of the errors at this point, which is good. 

The next step will be to write some tests for SPC and NCH. To avoid hitting live servers (which can go down and may not be reliable for testing purposes), we utilize `vcrpy` under the hood. Here is an example:

https://github.com/Unidata/siphon/blob/master/siphon/tests/test_wyoming.py

(the `vcrpy` library gets pulled in by `from siphon.testing import get_recorder`).

Basically, the first time you run the test code, any web request gets "recorded" by `vcrpy` in a "cassette" that will be played back each time you run the test in the future. These "cassettes" are stored in the github repository just like the code, and are found [here](https://github.com/Unidata/siphon/tree/master/siphon/tests/fixtures).

It's not super complicated, but it's a little tricky the first time you go through the testing processing like this.

zbruick:
I'll get my hands dirty with siphon and work on fixing the Travis errors.

AodhanSweeney:
Sorry guys just seeing this now but I will make sure to get on the tests.
Thanks for the heads up

zbruick:
Note that this PR will close #259 and #257.

zbruick:
The remaining Travis issue isn't caused by this PR (see discussion on #269). So as far as issues to fix, I think you're good for right now. I think this just needs to go to everyone (@lesserwhirls @dopplershift) for final review now.

AodhanSweeney:
I saw that it also wanted a copyright so i gave it that as well, I think this is as far as I can take it though... Thanks for all the help though man! @zbruick

siphon/simplewebservice/iastate.py
```diff
@@ -173,6 +173,6 @@ def _get_data_raw(self, time, site_id, pressure=None):
                 message += 'for station {stid}'.format(stid=site_id)
             if pressure is not None:
                 message += 'for pressure {pres}'.format(pres=pressure)
-            message = message[:-1] + '.'
+            message = message + '.'
```
lesserwhirls: @zbruick - should this be here, or is this something that accidentally made it in this PR during the merge?

siphon/simplewebservice/nhc.py
```diff
@@ -0,0 +1,163 @@
...
+            unformatted_forecast_url = 'http://ftp.nhc.noaa.gov/atcf/archive/{}/a{}.dat.gz'
+            urlf = unformatted_forecast_url.format(year, filename[1:])
+            unformatted_best_url = 'http://ftp.nhc.noaa.gov/atcf/archive/{}/b{}.dat.gz'
+            urlb = unformatted_best_url.format(year, filename[1:])
+
+        url_links = [urlf, urlb]
+        url_count = 0
+        for url in url_links:
+            # Checking if url is valid, if status_code is 200 then website is active
+            if requests.get(url).status_code == 200:
```
lesserwhirls: There might be cases where you get another value of 2xx code that is actually ok. Rather than checking for 200 exactly, you can use `raise_for_status()` on the response, and if something went wrong (4xx, 5xx), requests will raise a `requests.exceptions.HTTPError` error. See https://3.python-requests.org/user/quickstart/#response-status-codes.

You could replace the whole if block with something like:

~~~python
try:
    response = requests.get(url)
    response .raise_for_status()
    # rest of code after old if statement
    ...
except requests.exceptions.HTTPError as http_error:
    # give helpful message on console
    print("url {} was not valid, select different storm.".format(url))
    # raise the caught error
    raise    
~~~

lesserwhirls: Now I'm wondering...since there is only one direct use of the requests library in your PR, perhaps class `NHCD` should subclass `HTTPEndPoint`, much like `IAStateUpperAir` does. @dopplershift @zbruick? Or is that something we refactor later, given the summer timeline?

siphon/simplewebservice/spc.py
```diff
@@ -0,0 +1,143 @@
...
+Center's Archive of data that goes back to the 1950s.
+Weather events that are available are
+hail, wind, and tornados.
+
+"""
+
+import pandas as pd
+
+
+class SpcData:
```
lesserwhirls: `SpcData` should also use `HTTPEndPoint`, in the way that `NDBC` does. In this case, we would use the methods from `HTTPEndPoint` to read the data from a URL, and pass that to pandas `pd.read_csv` by wrapping the `HTTPEndPoint` return in a `StringIO`.  Again,  @dopplershift @zbruick, is that something we refactor later, given the summer timeline?

siphon/simplewebservice/iastate.py
AodhanSweeney: This, surprisingly, is a purposeful edit. It's in response to Error #257 in siphon about truncation of the final character in a string. I tested a few different scenarios and weird as it seems, the error seemed to be that the final character was continuously cut off.

lesserwhirls: Awesome, just wanted to be sure. Sometimes little things squeek through a rebase and conflict resolve session.

siphon/simplewebservice/nhc.py
zbruick: Yeah I think consistent methods for accessing data whenever possible across the package would be nice. Since I haven't played around with any of this, I'm not sure how much refactoring it would take, but given how `IAStateUpperAir` uses it, I don't think it would take that much time. @AodhanSweeney, chime in if this would take too long - not sure what's left on your to-do list for the summer. If too much, we can refactor down the road.

zbruick: We would still need to check if the URL was valid though, right? `NDBC` still does this with `_check_if_url_valid`, so I don't believe that's handled by `HTTPEndPoint`?

lesserwhirls: I can think of two things to do to address this.

1) replace all `r.status_code == 200` checks with `r.raise_for_status()` (when you have a response obtained from `HTTPEndPoint`)

2) extend `HTTPEndPoint` to have a method for checking if a url is reachable (`is_url_reachable(url)`?) which simply does a head request on the url followed by a `raise_for_status()` (would replace all `_check_if_url_valid` type things). Maybe something like:

~~~python
def is_url_reachable(self, path, params=None):
        """Test if a url is reachable.

        Make a HEAD request, optionally including a parameters, to a path.
        The path of the request is the full URL.
        Parameters
        ----------
        path : str
            The URL to request
        params : DataQuery, optional
            The query to pass when making the request
        Returns
        -------
        resp : requests.Response
            The server's response to the request
        Raises
        ------
        HTTPError
            If the server returns a 4xx or 5xx code

        """
        resp = self._session.head(path, params=params)
        resp.raise_for_status()

        return true
~~~

Since `get` requests follow redirect by default, we should probably make sure the `head` call includes the parameter `allow_redirects=True` (`head` does not follow redirects by default), since the url would ultimately be reachable by a `get` request.

How does that sound?

AodhanSweeney:
This most recent pull request is good for approval, but I want to work something else out so no one needs to look at it quite yet

AodhanSweeney:
This is now ready for review, once this is in I will be pushing respective GUIs into the python library.

AodhanSweeney:
This should be ready for review, it addresses issues #228 and #259

lesserwhirls:
The build isn't passing at the moment, but outside of that, what remains on this one?

dopplershift:
A review of the API for the NHC stuff (signature of public functions) to make sure we're ok with supporting that interface long term.

PR #272, opened 2019-07-13
Catalog strptime

deeplycloudy:
Some datasets, including various GOES datasets on the Unidata THREDDS catalog, don't have a time dimension, but that time dimension can be inferred from the filename. However, those GOES filenames have julian days in the timestamps, making it not possible to trivially parse them with the current regex approach in DatasetCollection.

This PR adds an additional regex tag `strptime` and an additional argument for supplying the `datetime.strptime` format string. This functionality increases the number of filenames from which time information can be harvested .


zbruick:
Looks like there are some flake8 issues and we probably need a test added for this as well.

deeplycloudy:
It looks like there is no existing test for `regex=` with `filter_time_*`, so do we really need two tests — one for the old behavior, and one for the new behavior with `strptime`?

zbruick:
Good catch - yeah that would probably be best.

deeplycloudy:
I added two tests: a standalone regex and regex with strptime. I don't understand the flake8 errors on travis, though I gather from above there may be some big-picture things driving those errors, and not this PR?

siphon/catalog.py
```diff
@@ -59,19 +60,25 @@ def _get_datasets_with_times(self, regex):
...
                 yield dt, self[ds]
 
         # If we never found any keys that match, we should let the user know that rather
         # than have it be the same as if nothing matched filters
         if not found_date:
             raise ValueError('No datasets with times found.')
 
-    def filter_time_nearest(self, time, regex=None):
+    def filter_time_nearest(self, time, regex=None, strptime=None):
         """Filter keys for an item closest to the desired time.
```
zbruick: To make flake8 happy with 2.7
```suggestion
        r"""Filter keys for an item closest to the desired time.
```

siphon/catalog.py
```diff
@@ -86,20 +93,27 @@ def filter_time_nearest(self, time, regex=None):
...
-            'second', and 'microsecond', as appropriate. When a match is found, any of those
-            groups missing from the pattern will be assigned a value of 0. The default pattern
-            looks for patterns like: 20171118_2356.
+            given, this should contain either
+            1. named groups: 'year', 'month', 'day', 'hour', 'minute', 'second',
+            and 'microsecond', as appropriate. When a match is found, any of those groups
+            missing from the pattern will be assigned a value of 0. The default pattern looks
+            for patterns like: 20171118_2356.
+            or
+            2. a group named 'strptime' (e.g., r'_s(?P<strptime>\d{13})' for GOES-16 data)
```
zbruick: Flake8 isn't happy about the escape `/d` here. I don't know if we need the explicit example in the docstring - if it is, we should add `# noqa:W605`.

siphon/catalog.py
```diff
@@ -86,20 +93,27 @@ def filter_time_nearest(self, time, regex=None):
...
             The value with a time closest to that desired
 
         """
-        return min(self._get_datasets_with_times(regex),
+        return min(self._get_datasets_with_times(regex, strptime),
                    key=lambda i: abs((i[0] - time).total_seconds()))[-1]
 
-    def filter_time_range(self, start, end, regex=None):
+    def filter_time_range(self, start, end, regex=None, strptime=None):
         """Filter keys for all items within the desired time range.
```
zbruick: Same as first comment
```suggestion
        r"""Filter keys for all items within the desired time range.
```

siphon/catalog.py
```diff
@@ -115,17 +129,24 @@ def filter_time_range(self, start, end, regex=None):
...
-            'second', and 'microsecond', as appropriate. When a match is found, any of those
-            groups missing from the pattern will be assigned a value of 0. The default pattern
-            looks for patterns like: 20171118_2356.
+            given, this should contain either
+            1. named groups: 'year', 'month', 'day', 'hour', 'minute', 'second',
+            and 'microsecond', as appropriate. When a match is found, any of those groups
+            missing from the pattern will be assigned a value of 0. The default pattern looks
+            for patterns like: 20171118_2356.
+            or
+            2. a group named 'strptime' (e.g., r'_s(?P<strptime>\d{13})' for GOES-16 data)
```
zbruick: Same issue about the escaping character

siphon/catalog.py
deeplycloudy: Fixed in commit just now.

siphon/catalog.py
deeplycloudy: Fixed in commit just now.

siphon/catalog.py
deeplycloudy: Kind of curious to see if the raw string fixes this.

I added an example because I like having something to start from so I don't have to look up the regex and strptime docs. Fine with me to match house style, of course.

dopplershift: I think it's fine to include, if for no other reason I think 0.001% of our potential audience would otherwise know how to write a regex with a named group without looking at docs. Count myself among the 99.999%.

We shouldn't avoid this in the name of making the tool happy.

siphon/tests/test_catalog.py
```diff
@@ -221,6 +221,46 @@ def test_datasets_time_range():
                       'NAM_CONUS_20km_noaaport_20150529_0000.grib1']
 
 
+@recorder.use_cassette('top_level_20km_rap_catalog')
+def test_datasets_time_range_regex():
+    """Test getting datasets by time range using filenames, with manual regex"""
```
dopplershift: Bot's likely to complain about the lack of the period at the end of the docstring.

siphon/tests/test_catalog.py
```diff
@@ -221,6 +221,46 @@ def test_datasets_time_range():
...
+    assert titles == ['NAM_CONUS_20km_noaaport_20150528_0000.grib1',
+                      'NAM_CONUS_20km_noaaport_20150528_0600.grib1',
+                      'NAM_CONUS_20km_noaaport_20150528_1200.grib1',
+                      'NAM_CONUS_20km_noaaport_20150528_1800.grib1',
+                      'NAM_CONUS_20km_noaaport_20150529_0000.grib1']
+
+
+@recorder.use_cassette('top_level_20km_rap_catalog')
+def test_datasets_time_range_strptime():
+    """Test getting datasets by time range using filenames, with strptime"""
```
dopplershift: Here too.

dopplershift:
It's been a couple weeks of annoying package incompatibilities. `pydocstyle` released 4.0, which breaks `flake8-docstrings`. 😭 I'll push a commit to your branch that should fix it.

dopplershift:
Well, that fixed the errors about `pydocstyle`.

zbruick:
Looks like there's just a couple lint issues left to deal with on this and then it's good to go!

siphon/tests/test_catalog.py
deeplycloudy: Fixed in 5fa3b9b

siphon/tests/test_catalog.py
deeplycloudy: Fixed in 5fa3b9b

deeplycloudy:
I think the remaining lint issues are handled. Looks like the only remaining error is a connection error in some part of the code not related to this PR?

zbruick:
Thanks for your work on this! So the remaining issue is documented in #273 and doesn't have anything to do with this PR. I think this is good to merge, but I'll defer to @dopplershift.

dopplershift:
Thanks @deeplycloudy !

deeplycloudy:
My pleasure, @dopplershift  and my thanks to you and @zbruick for a pleasant PR process!

PR #274, opened 2019-07-17
Refactor of IGRA2 data access from FTP to HTTP

zbruick:
Due to Travis not being able to hit the FTP server right, and NOAA dropping FTP support eventually, this refactor helps streamline our data access to HTTP. Right now, the tests pass, but take a long time due to the file size (total of 270 MB). This needs to be worked on with vcrpy before merging.


dopplershift:
I have some code that will fixup the fixtures. Can you fixup and rebase and I'll push the fix up?

zbruick:
Fixed and rebased. Go ahead with the fixtures

siphon/simplewebservice/igra2.py
```diff
@@ -102,15 +103,17 @@ def _get_data_raw(self):
...
-        except ImportError:
-            from urllib2 import urlopen
-
-        with closing(urlopen(self.ftpsite + self.site_id + self.suffix + '.zip')) as url:
-            f = ZipFile(BytesIO(url.read()), 'r').open(self.site_id + self.suffix)
+        path = self.folder + self.site_id + self.suffix + '.zip'
+
+        resp = self.get_path(path)
+        # See if the return is valid, but has no data
+        if resp.raise_for_status() is not None:
```
dopplershift: Not quite right. So if you look in the catalog code, we do:

https://github.com/Unidata/siphon/blob/b3c6f2de31702d3f00c3b8827f6278c039494144/siphon/catalog.py#L274-L281

What happens is that if an HTTP error was returned, it will raise an error, otherwise things continue fine. (See [here](https://2.python-requests.org/en/master/user/quickstart/#response-status-codes) for the docs.) So when you call this function, if there's an error, you'll never get to the code to get the ValueError. I think that `ValueError` with that message is more helpful to our users, so we should keep it. You'll need to wrap `raise_for_status()` in a `try...except` to handle `http_util.HTTPError`--then the exception handling can raise the new `ValueError`.

zbruick: Ah, ok that makes more sense. I'll fix that.

zbruick: Ok, let me know if that's better or not

dopplershift:
Can you review my stuff (make sure it makes sense, etc.)?

dopplershift:
That's what I get for pushing without double checking my last test. I had to modify it and no longer use `raise_for_status()` because our `get_path()` function was already handling that for us--so I moved that inside the `try` block.

zbruick:
This looks good to go now.

PR #276, opened 2019-08-15
Add issue and PR templates

dopplershift:
Based on MetPy's.


dopplershift:
This doesn't change any code, so I'm ignoring the test failures.

PR #277, opened 2019-08-21
More Maintenance

dopplershift:
#### Description Of Changes
A variety of little things:
* Add `python_requires` to `setup.py`, which will be good as the next release is the last for 2.7
* Update Travis config to no longer build Python 3.6-dev, but instead add doc build on 3.7-dev
* Enable a couple more flake8 plugins

This should fail initially with the new plugins, at which point I'll add a commit fixing the failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # After changing this file, check it on:
 #   http://lint.travis-ci.org/
 language: python
-sudo: falsei
+sudo: false
 dist: xenial
 
 addons:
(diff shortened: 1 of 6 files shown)


zbruick:
I don't think I have administrator privileges to merge over the codecov issue, which is spurious.

dopplershift:
Codecov actually isn't totally spurious--for some reason Travis runs aren't successfully uploading the coverage reports, and there are a few lines that are specific to windows in code base. I've also seen this on MetPy. I'll dig in...

dopplershift:
Oh, my favorite problem: spurious network failure. Looks like everything is fine now with just a re-run of the CI builds.

PR #283, opened 2019-09-30
Add warning if time range is invalid

### LINKED ISSUE #244: Time Range Warning
When getting data with a time range, we should warn if start is after end. Users can be confused by the empty list return if they don't realize they have done this.

zbruick:
#### Description Of Changes
Adds warning to time range methods in `Catalog` and `http_util` if the time range is invalid.

#### Checklist

- [x] Closes #244
- [x] Tests added


zbruick:
Also updated a buoy number in an example due to lack of water temperature data from that buoy. This PR fixes the current doc failures.

PR #284, opened 2019-10-01
Add improved reprs

### LINKED ISSUE #260: improve repr of siphon objects
I am trying to learn to use siphon. It seems very powerful. Thank you for creating it!

My usual method for learning a new library is to try to introspect the objects via a combination of doc strings, tab completion, etc (in addition to rtfm of course). One challenge with siphon is that it's `__repr__` methods for its objects are often insufficiently informative, making it hard to quickly understand the datatypes returned by the different methods. For example.

```python
>>> from siphon.catalog import TDSCatalog
>> cat_url = 'https://www.ncei.noaa.gov/thredds/catalog/avhrr-patmos-x-cloudprops-files/catalog.xml'
>> cat = TDSCatalog(cat_url)
>>> cat
No name found
```
I would have preferred to see
```
<siphon.catalog.TDSCatalog 'unnamed catalog'>
```

Or
```python
>>> for k, v in cat.catalog_refs.items():
>>>     print(k, v)
>>>     break
1979 1979
```
Here this is especially confusing because `k` really is a string but `v` is a `siphon.catalog.CatalogRef`. Again, a better repr would be
```
<siphon.catalog.CatalogRef '1979'>
```

I believe that making these reprs more verbose would make the library easier to use for new users.

zbruick:
#### Description Of Changes
Add improved reprs based on feedback in #260. 

#### Checklist

- [x] Closes #260
- [x] Tests added


siphon/catalog.py
```diff
@@ -36,6 +36,10 @@ def __getitem__(self, item):
         except TypeError:
             return list(self.values())[item]
 
+    def __repr__(self):
+        """Return useful repr."""
+        return IndexableMapping
```
dopplershift: You’re returning the class itself here, which I’m pretty sure is not what you meant.

siphon/catalog.py
```diff
@@ -393,7 +399,9 @@ def latest(self):
                 return TDSCatalog(latest_cat).datasets[0]
         raise AttributeError('"latest" not available for this catalog')
 
-    __repr__ = __str__
+    def __repr__(self):
+        """Return informative repr for TDSCatalog."""
+        return str(TDSCatalog) + ' ' + self.catalog_name
```
dopplershift: Isn’t this just returning then `self.catalog_name + ‘ ‘ + self.catalog_name`?

siphon/catalog.py
zbruick: I didn't mean to return a repr here...it was left over from me figuring out how this module worked. I removed it for now.

siphon/catalog.py
zbruick: Updated this to a new return

zbruick:
I can reproduce the error locally on the Radar example. I'll look into it.

zbruick:
Seems like the radar issue was spurious. This is ready for review again.

PR #287, opened 2019-10-08
Add flag for IAState to interpolate NaNs 

### LINKED ISSUE #285: Add flag for returning clean IAState UpperAir data
Add a flag to allow interpolation of NaNs in temperature and dewpoint due to additional pressure levels from the significant wind data. This will allow easier plotting of data by beginning users of this API.

zbruick:
#### Description Of Changes
Adds a flag to interpolate NaNs for temperature and dewpoint at significant wind levels. Default is False, so the default return does not change. To do this interpolation, it makes use of pandas `interpolate()`, which is linear. This saves the end user from figuring out how to do this with pandas themselves and allows for similar code to the Wyoming returns by just adding `interp_nans=True` to the `.request_data()` call.

#### Checklist

- [x] Closes #285
- [x] Tests added
- [x] Fully documented

### DIFF
--- a/siphon/simplewebservice/iastate.py
+++ b/siphon/simplewebservice/iastate.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2013-2018 Siphon Contributors.
+# Copyright (c) 2013-2019 Siphon Contributors.
 # Distributed under the terms of the BSD 3-Clause License.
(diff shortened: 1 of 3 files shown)


dopplershift:
Needs a rebase on master to fix CI.

zbruick:
Rebased and pushed.

PR #293, opened 2019-11-15
Include tests in PyPI tarball

kmosiejczuk:
In OpenBSD we use the regression tests in order to make sure that updates
work properly and that an update of a dependency doesn't break a package.
Having the regression tests in the PyPI tarball makes that much easier.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,6 +5,7 @@ include versioneer.py
 include siphon/_version.py
 recursive-include examples *
 recursive-include docs *
+recursive-include siphon/tests *
 prune docs/build
 prune docs/examples
 prune docs/api/generated


dopplershift:
Thanks for the contribution. I'm actually surprised these aren't already included.

To be clear, you only need these included in the source distribution, correct? We're just looking soon at redoing the repository layout, and I want to make sure I keep the right use case in mind.

kmosiejczuk:
> Thanks for the contribution. I'm actually surprised these aren't already included.

> To be clear, you only need these included in the source distribution, correct? We're just looking soon at redoing the repository layout, and I want to make sure I keep the right use case in mind.

Correct. Just the source distribution.

PR #300, opened 2020-02-19
Add improved reprs for TDSCatalog

dcamron:
Starting changes based off @zbruick repr branch from #284


lesserwhirls:
Thanks @dcamron! The failures do not appear to be related to the changes here, but we can look into them tomorrow, as this is a good opportunity to take care of those failures. As I mentioned earlier today, there are a few additions here that would be nice, but we can discuss that more tomorrow (assuming the snow does not tank the day!)

PR #308, opened 2020-07-16
MesoWest addition to simplewebservice

PharaohCola13:
### Description Of Changes
Adds the [University of Utah's MesoWest](https://mesowest.utah.edu/) archive to the capabilities of siphon simplewebservice. The functions are written to dynamically support the irregularities and inconsistencies of the archive.

#### Checklist
- [x] Tests added
- [x] Fully documented

### DIFF
--- a/siphon/simplewebservice/mesowest.py
+++ b/siphon/simplewebservice/mesowest.py
@@ -1,0 +1,123 @@
+from datetime import datetime
+from io import StringIO
+import warnings, sys
+
+from bs4 import BeautifulSoup
+import numpy as np
+import pandas as pd
+
+from ..http_util import HTTPEndPoint
+
+warnings.simplefilter("ignore")
+
+class MesoWest(HTTPEndPoint):
+    """Download and parse data from the University of Utah's MesoWest archive."""
+    def __init__(self):
+        """Set up endpoint."""
+        super(MesoWest, self).__init__('https://mesowest.utah.edu/cgi-bin/droman/')
+
+    @classmethod
+    def request_data(cls, date, site_id, **kwargs):
+        r"""Retrieve upper air observations from the University of Utah MesoWest archive.
+
+        Parameters
+        ----------
+        date : datetime
+            The date of the desired observation.
+
+        site_id : str
+            The four letter MesoWest identifier of the station for which data should be
+            downloaded.
+            https://mesowest.utah.edu/cgi-bin/droman/meso_station.cgi?area=1
+
+        kwargs
+            Arbitrary keyword arguments to use to initialize source
+
+        Returns
+        -------
+            :class:`pandas.DataFrame` containing the data
+
+        """
+        endpoint = cls()
+        df = endpoint._get_data(date, site_id)
+        return df
+
+    def _get_data(self, date, site_id):
+        """Download and parse upper air observations from an online archive.
+
+        Parameters
+        ----------
+        date : datetime
+            The date of the desired observation.
+
+        site_id : str
+            The four letter MesoWest identifier of the station for which data should be
+            downloaded.
+            https://mesowest.utah.edu/cgi-bin/droman/meso_station.cgi?area=1
+
+        Returns
+        -------
+            :class:`pandas.DataFrame` containing the data
+
+        """
+        raw_data = self._get_data_raw(date, site_id)
+        soup = BeautifulSoup(raw_data, 'html.parser')
+
+        names = pd.DataFrame.from_records([[td.find_next(text=True).strip('\n\t\t') for td in tr.find_all('small')] for tr in soup.find_all('th')])[1].dropna(how='any', axis=0).reset_index(drop=True)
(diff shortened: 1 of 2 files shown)


siphon/simplewebservice/mesowest.py
```diff
@@ -0,0 +1,124 @@
+import warnings
+
+from bs4 import BeautifulSoup
+import numpy as np
+import pandas as pd
+
+from ..http_util import HTTPEndPoint
+
+warnings.simplefilter("ignore")
```
dopplershift, line 9: What warnings are we needing to silence?

siphon/simplewebservice/mesowest.py
```diff
@@ -0,0 +1,124 @@
...
+
+class MesoWest(HTTPEndPoint):
+    """Download and parse data from the University of Utah's MesoWest archive."""
+    def __init__(self):
+        """Set up endpoint."""
+        super(MesoWest, self).__init__('https://mesowest.utah.edu/cgi-bin/droman/')
```
dopplershift, line 15: Using this as the API endpoint seems...sketchy? Can you point to somewhere that says this is a good endpoint to use?

siphon/simplewebservice/mesowest.py
```diff
@@ -0,0 +1,124 @@
...
+        df[0] = pd.to_datetime(df[0]).apply(lambda x: x.time())
+
+        name = []
+        for i in range(1, len(df.columns)):
+            if str(df[i].dtypes) == 'object':
+                if str(df[i].iloc[0]).replace(' ', '').isalpha():
+                    pass
+                else:
+                    for j in range(0, len(df)):
+                        if '<' in str(df[i].iloc[j]).replace(' ', ''):
```
dopplershift, line 80: If you're just checking for the presence of '>' and '<', you don't need to call `replace`. I think the whole block could be:

```python
if not ('<' in str(df[i].iloc[j]) or '>' in str(df[i].iloc[j])):
    df[i].iloc[j] = float(df[i].iloc[j])
```

dopplershift:
@joeyoun9 Thoughts on the use of https://mesowest.utah.edu/cgi-bin/droman/ URL here for accessing data?

siphon/simplewebservice/mesowest.py
PharaohCola13: The warning that is being filtered is SettingWithCopyWarning. It is a consequence of rewriting the DataFrame Series slices into their proper type in the loop starting at line 74.

siphon/simplewebservice/mesowest.py
PharaohCola13: It is a bit odd that the archive it setup like this. I am running some tests to see if the https://mesowest.utah.edu/ endpoint works, but the cgi-bin/droman/ part will likely end up moved down to the path variable in the  _get_data_raw function.

siphon/simplewebservice/mesowest.py
PharaohCola13: This seems like a much better method of handling the < and > symbols that sometimes occur in the visibility data. This will be implemented in the next commit.

joeyoun9:
Hey all, 

It looks like this is some kind of a web scraper concept coming together here? Unfortunately though the present iteration of MesoWest may technically make that possible, it is unfortunately not something we support in any form. Historically we have been compelled to block sources of significant scraping requests. In time the entire MesoWest website is going to transition to a static HTML + APIs platform, at which point what I think this might be would not work.

So, no, I have to advise against this.

The data visible on MesoWest are available for programmatic access via Synoptic Data PBC (developers.synopticdata.com), however as a commercial service there is no unrestricted free use functionality for their ("our", I work there too) data access services (at the moment).

wxguy:
> 
> 
> Hey all,
> 
> It looks like this is some kind of a web scraper concept coming together here? Unfortunately though the present iteration of MesoWest may technically make that possible, it is unfortunately not something we support in any form. Historically we have been compelled to block sources of significant scraping requests. In time the entire MesoWest website is going to transition to a static HTML + APIs platform, at which point what I think this might be would not work.
> 
> So, no, I have to advise against this.
> 
> The data visible on MesoWest are available for programmatic access via Synoptic Data PBC (developers.synopticdata.com), however as a commercial service there is no unrestricted free use functionality for their ("our", I work there too) data access services (at the moment).

Can you point to the guide/ section where it is possible to download upper air data from developers.synopticdata.com. I have been reading and trying a lot since last two hours to get it working. But it appear to be no support for upper air data from developers.synopticdata.com.

Writing few words here about how to reterive such data set would help others as well. I mainly came here with a hope to plot upper air wind plot as shown here https://unidata.github.io/MetPy/dev/examples/plots/upperair_declarative.html#sphx-glr-examples-plots-upperair-declarative-py for west Asia region using MesoWest. After making hands dirty, it is understood that the data is not available to countries other than north America.

PR #321, opened 2021-02-11
fixed rhgrad import issue

DanielWatkins:
All the was missing was the column name, column units, and column conversion function for the column "relative_humidity_gradient" in the igra2-derived import function.


dopplershift (approved):
Can check that tests still pass when I fix CI.

PR #362, opened 2021-04-08
Expand documentation on indexing and interation for DatasetCollection

### LINKED ISSUE #361: Types not the same for indexing vs. iteration of catalog.datasets
Try to include:

- Code Sample: here the URL doesn't matter, but is provided for convenience.

```python
from datetime import datetime, timedelta
from siphon.catalog import TDSCatalog

day = datetime(2018, 5, 24)
start = day + timedelta(hours=16)
end = day + timedelta(hours=24)
url = f"https://data.nssl.noaa.gov/thredds/catalog/WRDD/OKLMA/{day:%Y/%m/%d}/catalog.xml"

cat = TDSCatalog(url)

print(type(cat.datasets))
# <class 'siphon.catalog.DatasetCollection'>

print(type(cat.datasets[0]))
# <class 'siphon.catalog.Dataset'>

for ds in cat.datasets:
    print(type(ds))
#<class 'str'>
#<class 'str'>
# etc.
```

- Problem description: When indexing `cat.datasets`, the return type is a Dataset object, but when iterating the return type is a string.
- Expected output: I expected the same return type, preferably a Dataset object. It would seem to be more Pythonic if DatasetCollection behaved like a list or tuple of objects. 
- Background: Since this catalog does not index the time of these data, I had to write a manual loop to parse the time from the dataset name. I used the indexing method to figure out what was in `cat.datasets`, and developed a parsing method based on that return type. That information was inaccurate when I converted the logic to a function to loop over datasets.
- Which platform: Mac
- Versions. Include the output of:
  - `python --version`: Python 3.6.12
  - `python -c 'import siphon; print(siphon.__version__)': 0.9

deeplycloudy:
#### Description Of Changes

Closes #361 with a documentation-update.

#### Checklist

- [x] Closes #361

### DIFF
--- a/src/siphon/catalog.py
+++ b/src/siphon/catalog.py
@@ -35,7 +35,19 @@ class IndexableMapping(OrderedDict):
 
 
 class DatasetCollection(IndexableMapping):
-    """Extend ``IndexableMapping`` to allow datetime-based filter queries."""
+    r"""Extend ``IndexableMapping`` to allow datetime-based filter queries.
+
+    Indexing works like a dictionary. The dataset name ('my_data.nc', a string) is the key,
+    and the value returned is an instance of ``Dataset``. Positional indexing
+    (e.g., [0]) is another valid method of indexing.
+
+    ``DatasetCollection`` is commonly encountered as the ``datasets`` attribute of a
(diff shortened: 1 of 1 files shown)


dopplershift (dismissed):
Nice improvement to the docs to finally get in.

dopplershift (approved):

PR #450, opened 2021-11-17
Update Python versions

### LINKED ISSUE #214: Dropping support for Python 2.7
So given that the ecosystem is [dropping support for 2.7 in earnest](http://python3statement.org), the time has come for us to start planning for Siphon to move on from Legacy Python:

- Core Python developers will [stop support for Python 2.7 January 1, 2020](https://pythonclock.org)
- NumPy feature releases will be [Python 3 only starting January 1, 2019](https://docs.scipy.org/doc/numpy/neps/dropping-python2.7-proposal.html), and support for the last release supporting Python 2 will end January 1, 2020.
- XArray will drop [2.7 January 1, 2019 as well](https://github.com/pydata/xarray/issues/1830)
- Matplotlib's 3.0 release, tentatively Summer 2018, [will be Python 3 only](https://mail.python.org/pipermail/matplotlib-devel/2017-October/000892.html); the current 2.2 release will be the last long term release that supports 2.7.

Given these events, and based on preliminary input from Unidata's Users' committee, I propose that Siphon drop support for Python 2.7 in the Fall of 2019. That implies that the Fall 2019 release of Siphon will be the first that supports only Python 3. The release before that (Spring or Summer 2019) will be the last that supports Python 2.7; given Siphon's relatively nascent status, I also propose that this release NOT have any particular long term support status.

I imagine this will impact some of our users, so I welcome feedback here. Note that we don't have the capacity to maintain all of Siphon's upstream dependencies, so some of this transition is beyond our control.

dopplershift:
#### Description Of Changes
* Finally drop 2.7 (#214)
* Drop Python 3.6
* Update CI a bit

#### Checklist

- [x] Closes #214 
~- [ ] Tests added~
- [x] Fully documented

### DIFF
--- a/src/siphon/cdmr/ncstream.py
+++ b/src/siphon/cdmr/ncstream.py
@@ -3,8 +3,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """Handle binary stream returns in NCStream format."""
 
-from __future__ import print_function
-
 from collections import OrderedDict
 import itertools
 import logging
@@ -62,8 +60,7 @@ def read_ncstream_data(fobj):
             bin_data = zlib.decompress(bin_data)
             assert len(bin_data) == data.uncompressedSize
         elif data.compress != stream.NONE:
-            raise NotImplementedError('Compression type {0} not implemented!'.format(
-                data.compress))
+            raise NotImplementedError(f'Compression type {data.compress} not implemented!')
 
         # Turn bytes into an array
         return reshape_array(data, np.frombuffer(bin_data, dtype=dt))
@@ -87,8 +84,7 @@ def read_ncstream_data(fobj):
             magic = read_magic(fobj)
         return data, blocks
     else:
-        raise NotImplementedError("Don't know how to handle data type: {0}".format(
-            data.dataType))
+        raise NotImplementedError(f"Don't know how to handle data type: {data.dataType}")
 
 
 def read_ncstream_data2(fobj):
@@ -144,7 +140,7 @@ def read_messages(fobj, magic_table):
         if func is not None:
             messages.append(func(fobj))
         else:
-            log.error('Unknown magic: ' + str(' '.join('{0:02x}'.format(b)
+            log.error('Unknown magic: ' + str(' '.join(f'{b:02x}'
                                                        for b in bytearray(magic))))
 
     return messages
--- a/src/siphon/metadata.py
+++ b/src/siphon/metadata.py
@@ -3,8 +3,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """Helps support reading and parsing metadata elements from a TDS client catalog."""
 
-from __future__ import print_function
-
 import logging
 
 logging.basicConfig(level=logging.ERROR)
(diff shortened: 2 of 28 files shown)


src/siphon/cdmr/ncstream.py
```diff
@@ -87,8 +84,7 @@ def read_ncstream_data(fobj):
             magic = read_magic(fobj)
         return data, blocks
     else:
-        raise NotImplementedError("Don't know how to handle data type: {0}".format(
-            data.dataType))
+        raise NotImplementedError(f"Don't know how to handle data type: {data.dataType}")
```
dcamron, line 87:
```suggestion
        raise NotImplementedError(f'Don't know how to handle data type: {data.dataType}')
```
Random double quotes? Didn't check for any more of these.

src/siphon/metadata.py
```diff
@@ -513,7 +511,7 @@ def _get_handler(self, handler_name):
         elif handler_name in self._sts:
             return getattr(self._st, handler_name)
         else:
-            msg = 'cannot find handler for element {}'.format(handler_name)
+            msg = f'cannot find handler for element {handler_name}'
             log.warning(msg)
```
dcamron, line 515: Want to go ahead and flake8-pieify this as a log message (didn't dig for more outside of changes) while you're in here?

src/siphon/ncss_dataset.py
```diff
@@ -201,11 +200,11 @@ def lookup(self, handler_name):
         if handler_name in dir(self):
             return getattr(self, handler_name)
         else:
-            msg = 'cannot find handler for element {}'.format(handler_name)
+            msg = f'cannot find handler for element {handler_name}'
             log.warning(msg)
```
dcamron, line 204: Same question as `metadata.py:513` above.

dcamron (approved):
These are all just linting, which we can spare for some other style cleanup if preferred. I'll approve and leave that call up to you

src/siphon/cdmr/ncstream.py
dopplershift: Confused me for a sec just now, but the `'` in "Don't" allows this as double to avoid escaping it.

dcamron: Haha, totally missed it! Even with Github syntax highlighting pointing it out to me.

src/siphon/metadata.py
dopplershift: Sure. Did yours catch this? Because mine didn't.

src/siphon/ncss_dataset.py
dopplershift: Done.

dopplershift:
Thanks for catching those. I just added the logging changes as a separate commit, so re-review should be simple.

src/siphon/metadata.py
dcamron: I didn't run it myself, I just remembered making that change over on MetPy. I can check and see if it does.

dcamron: Update: it doesn't.

dcamron (approved):

dopplershift:
CodeQL is succeeding, but for some reason is failing to report. 🤷‍♂️

PR #453, opened 2021-11-18
MNT: move setup.py content into setup.cfg

### LINKED ISSUE #301: Update build system
Pull in the same things we do in MetPy:
- [x] Replace `versioneer` with `setuptools_scm`
- [x] Update `MANIFEST.in`
- [ ] Put everything we can in `setup.cfg` and strip down `setup.py`
- [x] Switch to having siphon source in `src/siphon`
- [x] Move tests to `tests/`, along with vcrpy cassettes
- [x] Source distributions (`python setup.py sdist`) should include tests and fixtures, wheels should not

akrherz:
#### Description Of Changes

Copy mostly-verbatim `setup.py` content into `setup.cfg`.

#### Checklist

- [x] Closes #301
- ~~Tests added~~
- ~~Fully documented~~

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,76 @@
+[metadata]
+name = siphon
+description = A collection of Python utilities for interacting with the Unidata technology stack.
+description_file = README.rst
+author = Unidata Development Team
+author_email = support-python@unidata.ucar.edu
+maintainer = Siphon Developers
+maintainer_email = support-python@unidata.ucar.edu
+license = BSD 3-Clause
+license_file = LICENSE
+platform = any
+keywords = meteorology, weather
+classifiers =
+    Development Status :: 3 - Alpha
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Topic :: Scientific/Engineering
+    Intended Audience :: Science/Research
+    Operating System :: OS Independent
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -1,3 +1,76 @@
+[metadata]
+name = siphon
+description = A collection of Python utilities for interacting with the Unidata technology stack.
+description-file = README.rst
```
dopplershift, line 4: While a copy/paste of what was there before, I'm pretty sure this needs to be `_` rather than `-`:
```suggestion
description_file = README.rst
```

dopplershift (changes requested):
Thanks for handling this maintenance. Just a minor tweak here.

setup.cfg
akrherz: Ah, thank you.

dopplershift:
Can you rebase this on top of the changes I made to `setup.py` with dropping Python 2?

setup.py
```diff
@@ -3,12 +3,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """Setup script for installing Siphon."""
 
-from __future__ import print_function
-
+from setuptools import setup
```
dopplershift, line 6: This needs to go below `import sys`:


```python
import sys

from setuptools import setup
```

akrherz: oopsy, senior moment there.

dopplershift (approved):

dopplershift:
CodeQL is passing but failing to report, so just going ahead and merging. Thanks again @akrherz !

PR #696, opened 2023-05-23
MNT: Fix WyomingUpperAir height folding

dcamron:
Pandas `read_fwf` only infers column widths based on the first 100 rows, leading to cut off values in later rows. Define column widths explicitly fixes this. Cut off after 8 columns since we ignore those anyway.

Thanks to the issue report in the python-users unidata mailing list and credit to @kgoebber for similar solution and test changes/additions.

#### Checklist

- [X] Tests added


dopplershift (approved):
Looks good. Failures look completely unrelated, so while they could use fixing, that will have to wait.

PR #755, opened 2024-01-08
Catalogue walk

### LINKED ISSUE #754: Add a walk function for navigating THREDDS catalogue
As far a I know this functionality does not exist already but believe it would be a welcome addition :

I often need to find all datasets for multiple subfolders of a thredds catalogue. To do this I resort to using a custom function in my data processing scripts (see simple example below) but ideally this would be built into `siphon` itself.

```
from siphon.catalog import TDSCatalog
# walk function
def walk(cat, depth=1):
    """Return a generator walking a THREDDS data catalog for datasets.

    Parameters
    ----------
    cat : TDSCatalog
      THREDDS catalog.
    depth : int
      Maximum recursive depth. Setting 0 will return only datasets within the top-level catalog. If None,
      depth is set to 1000.
    """
    yield from cat.datasets.items()
    if depth is None:
        depth = 1000

    if depth > 0:
        for name, ref in cat.catalog_refs.items():
            try:
                child = ref.follow()
                yield from walk(child, depth=depth - 1)

            except requests.HTTPError as exc:
                LOGGER.exception(exc)

# creat catalogue
cat = TDSCatalog(urlcat)
# access all dataset to 20 subfolders
for dd in (cat, depth=20):
    print(dd)

```

tlogan2000:
#### Description Of Changes
Implements a catalog walk method the the TDSCatalog class

#### Checklist

- [x] Closes #754 
- [x] Tests added
- [ ] Fully documented

### DIFF
--- a/src/siphon/catalog.py
+++ b/src/siphon/catalog.py
@@ -11,6 +11,7 @@ from collections import OrderedDict
 from datetime import datetime
 import logging
 import re
+import requests
 import warnings
 import xml.etree.ElementTree as ET  # noqa:N814
 try:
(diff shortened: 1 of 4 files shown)


src/siphon/catalog.py
```diff
@@ -389,6 +390,31 @@ def _process_datasets(self):
...
+        """Return a generator walking a THREDDS data catalog for datasets.
+
+        Parameters
+        ----------
+        cat : TDSCatalog
+          THREDDS catalog.
+        depth : int
+          Maximum recursive depth. Setting 0 will return only datasets within the top-level catalog. If None,
+          depth is set to 1000.
+        """
```
huard, line 404:
```suggestion

Yields
------
Dataset
    Siphon catalog dataset.



        """
```

tlogan2000:
@dopplershift   I'm not sure I understand why CI build are failing.   Failures for `test_catalog_walk3` and `test_catalog_walk0` are passing locally for me but admittedly this is the first time trying the `@reecorder.use_cassette()` decorator.  Happy to try to fix if you could give some guidance.

PR #805, opened 2024-10-31
Drop in MetPy docs workflow

dcamron:
First pass dropping in MetPy docs workflows as-is before potentially updating upstream actions


dopplershift:
I wonder if we even need scipy at all, since we're not doing any complicated remapping.

dcamron:
Assuming doc builds come back green this should be good to go @dopplershift

dopplershift (approved):

PR #810, opened 2024-11-04
Update linting workflow

dopplershift:
#### Description Of Changes
* Switch over to ruff for most of linting (matching MetPy)
* Migrate most of config over to `pyproject.toml`
* Add codespell to tools


dcamron (commented):
It looks like `developerguide.rst` needs a few more updates than just those included here. We can open a separate issue for that though.

dcamron (approved):

PR #811, opened 2024-11-05
Bump protobuf to 5.28.2

dopplershift:
#### Description Of Changes
* Bumps ci version to 5.28.2 (since 5.28.3 isn't available on conda-forge yet)


dopplershift:
Ok, looks like 5 breaks, so we can't rebuild with that yet. Just bumping latest to 5 but not rebuilding then.

PR #748, opened 2023-11-29
Fix WyomingUpperAir Missing Station Information bug

tkschuler:
#### Description Of Changes
Fixed bug #749 which happens with some soundings that are missing Latitude, Longitude, and Elevation data. 

Example:
`df = WyomingUpperAir.request_data(datetime(2012, 1, 1, 0),  '82244')`

#### Checklist
- [X] Tests added

PR #853, opened 2024-12-11
Make parsing Wyoming post-data values more robust

### LINKED ISSUE #749: WyomingUpperAir Missing Station Information
wyoming.py fails with:

`df = WyomingUpperAir.request_data(datetime(2012, 1, 1, 0),  '82244')`

This is because with the raw text on the server, Latitude is listed as '******'  and can't be converted to a float.  On further inspection,  Longitude is missing,  and Elevation is -9999.0  I've noticed this with multiple soundings in South America. 

I think it would make the most sense to return None or Nan for Latitude, Longitude and Elevation when this problem arises.

dopplershift:
The main goal is to allow for missing station information, which results in '******' for latitude, no longitude, and -9999.0 for elevation. As a result, avoid the fixed index parsing of this information, and instead rely on the text to the left of the ':' to identify appropriate sections.

#### Checklist

- [x] Fixes #749
- [x] Tests added

Supersedes and close #748.


tests/test_wyoming.py
```diff
@@ -140,3 +141,14 @@ def test_wyoming_heights():
...
     assert_almost_equal(df['direction'][1], 145.0, 1)
+
+
+# GH #749
+@recorder.use_cassette('wyoming_missing_station_info')
+def test_missing_station():
+    """Test that we can still return data for stations missing from the Wyoming archive."""
+    df = WyomingUpperAir.request_data(datetime(2012, 1, 1, 0),  '82244')
+    assert df['station'][0] == ''
+    assert pd.isna(df['latitude'][0])
```
lesserwhirls, line 152: Would it be worth using `all()` to make sure all values for these variables satisfy `pd.isna`?

dopplershift: Ah, good idea.

dopplershift: Updated.

lesserwhirls (approved):

PR #944, opened 2025-06-27
Update dependabot config

dopplershift:
#### Description Of Changes
Bring in changes from MetPy, including updating locations of linting, test, and doc requirements to be compatible with the actions from MetPy we use. Also move netcdf4 to extras.


.github/workflows/tests-pypi.yml
```diff
@@ -58,15 +58,16 @@ jobs:
...
+                           (opt_deps['extras'], 'ci/extra_requirements.txt')]:
+          with Path(fname).open('wt') as out:
             for dep in src:
               dep = dep.split(';')[0]
               if 'siphon' not in dep:
                 out.write(dep.replace('>=', '==') + '\n')
             # Needed until minium vcrpy is >=7.0.0 for urllib3>=2.3
-            if fname == 'requirements.txt':
+            if fname == 'ci/requirements.txt':
               out.write('urllib3==2.2.3\n')
```
lesserwhirls, line 71: While not really a part of the change review, this gives me the heebie-jeebies...does this only apply for GitHub actions? (won't hold up the PR for this, but asking out of...stomach pain...)

lesserwhirls (approved):

.github/workflows/tests-pypi.yml
dopplershift: The urllib3 hard-coded version?

PR #961, opened 2025-09-10
Replace flake8-continuation

dcamron:
flake8-broken-line gets us the same functionality and is maintained at present. ref Unidata/MetPy#3868


dopplershift:
Just need to clean up the `\` that are  in some tests.

PR #1066, opened 2026-07-15
Update WyomingUpperAir client

dopplershift:
#### Description Of Changes

Updates to work with new University of Wyoming server. Not only did they tweak the supported web API, but they have now added direct CSV access. This simplifies our parsing greatly, eliminates our need for BeautifulSoup, at the cost of losing some pre-calculated sounding indices (though it adds some more columns too).

I also refactored the basic test so we have one test that includes the `recalc` parameter. Currently this does nothing on the new server, but it's possible it may use it in the future.


src/siphon/simplewebservice/wyoming.py
```diff
@@ -155,17 +131,11 @@ def _get_data_raw(self, time, site_id, recalc=False):
...
-        resp = self.get_path(path)
-        # See if the return is valid, but has no data
-        if resp.text.find("Can't") != -1:
-            raise ValueError(
-                f'No data available for {time:%Y-%m-%d %HZ} '
-                f'for station {site_id}.')
-
-        return resp.text
+        try:
+            return self.get_path(path).text
```
lesserwhirls, line 139: Would it be worth checking that the response `Content-Type` header is `text/csv` here, and bailing out earlier than the pandas `read_csv` call if not? It might make it easier to diagnose server error or changes.

dopplershift: On error we should be triggering a 404 and exception now, so we don't get to Pandas? Previously we were scraping the message here and raising. Or what am I missing?

lesserwhirls: Since I don't know their server stack super well, it's possible a meaningful HTML message would be returned while still giving a 2xx response. In that case, we'd end up sending Pandas the HTML. That would result in a pandas error, possibly obscuring the underlying issue (that CSV wasn't returned to begin with). This is more from a defensive position than experience with the new server.

dopplershift: Reasonable, but one challenge: I can't figure out how to get the server to return a 200 for an error. So that error handling clause would be untested...

lesserwhirls: True. I've seen 200 responses abused so much in the past that it feels almost normal. But it also does not need to be an error necessarily. We can focus on the happy path for now, and deal with any special things the server throws at us if it actually does.

lesserwhirls (approved):

src/siphon/simplewebservice/wyoming.py
dcamron: Is it more broadly useful for siphon if we write some test helpers here to fake responses in addition/replacement to cassettes?

dopplershift: I'm not sure it's useful to write helpers to generate hypothetical cases--if we're faking it, there's no guarantee the code that handles the error case will actually function correctly in the real world. I think I'm with @lesserwhirls: let's wait until there's actually a problem. We need to just deal with what is, not what might be someday.

PR #1073, opened 2026-07-20
Fix IGRA release time parsing

### LINKED ISSUE #982: IGRAUpperAir returns wrong release time in the header due to excess padding of string in _ctime function
The _ctime function in igra2.py is supposed to work for both MMMSS and HHMM formats, but is always padding the input string to 5 characters, which causes the hour and minute substrings of the HHMM format to be wrong. This can be seen by running the following copy of the _ctime code, with some print statements added.:

```
# Copy of _ctime function from igra2.py, with print statements added
def _ctime(strformat='MMMSS'):
    """Return a function converting a string from MMMSS or HHMM to seconds."""
    def _ctime_strformat(val):
        print('val',val)
        time = val.strip().zfill(5)
        print('time',time)
        if int(time) < 0 or int(time) == 9999:
            return np.nan
        else:
            if strformat == 'MMMSS':
                minutes = int(time[0:3])
                seconds = int(time[3:5])
                time_seconds = minutes * 60 + seconds
            elif strformat == 'HHMM':
                hours = int(time[0:2])
                minutes = int(time[2:4])
                print('hours minutes', hours,minutes)
                time_seconds = hours * 3600 + minutes * 60
            else:
                sys.exit('Unrecognized time format')

        return time_seconds
    return _ctime_strformat

# Call the function and print the result
func = _ctime(strformat='HHMM')
print(func('1142'))

```

The result is this:

val 1142
time 01142
hours minutes 1 14
4440

You can see the hours and minutes are not being correctly extracted from the input, so the resulting number of seconds computed is wrong. To fix this, the 5-char padding should be applied only when the format is MMMSS, and 4-char padding should be applied when the format is HHMM.

Sanjays2402:
#### Description Of Changes

Closes #982. IGRA release times in `HHMM` format now use their four-character field width, preserving both hour digits. The regression covers `1142` converting to 11:42 rather than 01:14.

#### Checklist

- [x] Closes #982
- [x] Tests added

### DIFF
--- a/src/siphon/simplewebservice/igra2.py
+++ b/src/siphon/simplewebservice/igra2.py
@@ -6,7 +6,6 @@
 import datetime
 from io import BytesIO, StringIO
 import itertools
-import sys
 import warnings
 from zipfile import ZipFile
 
@@ -215,7 +214,7 @@ class IGRAUpperAir(HTTPEndPoint):
                         minutes = int(time[2:4])
                         time_seconds = hours * 3600 + minutes * 60
                     else:
-                        sys.exit('Unrecognized time format')
+                        raise ValueError(f'Unrecognized time format "{strformat}"')
 
                 return time_seconds
             return _ctime_strformat
--- a/src/siphon/simplewebservice/igra2.py
+++ b/src/siphon/simplewebservice/igra2.py
@@ -201,7 +201,7 @@ class IGRAUpperAir(HTTPEndPoint):
         def _ctime(strformat='MMMSS'):
             """Return a function converting a string from MMMSS or HHMM to seconds."""
             def _ctime_strformat(val):
-                time = val.strip().zfill(5)
+                time = val.strip().zfill(5 if strformat == 'MMMSS' else 4)
 
                 if int(time) < 0 or int(time) == 9999:
                     return np.nan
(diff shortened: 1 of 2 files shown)


src/siphon/simplewebservice/igra2.py
```diff
@@ -201,7 +201,7 @@ def _cflag(val):
         def _ctime(strformat='MMMSS'):
             """Return a function converting a string from MMMSS or HHMM to seconds."""
             def _ctime_strformat(val):
-                time = val.strip().zfill(5)
+                time = val.strip().zfill(5 if strformat == 'MMMSS' else 4)
```
dopplershift: How about:

```suggestion
                time = val.strip().zfill(len(strformat))
```

src/siphon/simplewebservice/igra2.py
dopplershift: While here, can you also replace this with:
```suggestion
                        raise ValueError(f'Unrecognized time format "{strformat}"')
```

and also remove the `import sys` at the top?

Sanjays2402:
Good call — switched it to raise ValueError(f'Unrecognized time format "{strformat}"') so it surfaces the offending format string, and dropped the now-unused import sys (it was only used by the old sys.exit). IGRA tests still green. Thanks!

src/siphon/simplewebservice/igra2.py
Sanjays2402: Applied — `zfill(len(strformat))` in 9be5b1b, which reads better than my conditional and stays right if another width shows up.

Sanity check on the HHMM path: `1142` -> 42120, `" 942"` -> 34920, `9999` -> nan. tests/test_igra2.py is 5 passed.

PR #1077, opened 2026-07-27
Patch out netCDF4.Dataset in test

### LINKED ISSUE #1076: Fix external server in test
Currently `test_oceandata_hyrax_dataset` in `test_catalog.py` relies on an external URL: `https://oceandata.sci.gsfc.nasa.gov/opendap/hyrax/SeaWiFS/L3SMI/2000/0101/SEASTAR_SEAWIFS_GAC.20000101.L3m.DAY.CHL.chlor_a.9km.nc`. This is because the URL gets read through the OPeNDAP support in netcdf4-python, and avoids the mechanisms of vcrpy.

This is now failing regularly, producing a bunch of false negatives. We should update the test to instead validate the assembled URL and ensure the right handler is called--but actual traffic using OPeNDAP isn't something siphon needs to test here.

dopplershift:
#### Description Of Changes
netCDF4-python OPeNDAP support isn't captured by vcrpy, so this test was accessing the server on every run. This is failing a bunch lately, so patch out the call to Dataset() and just validate that it's used properly--we don't need to check OPeNDAP traffic end-to-end.

This should greatly help more PRs run cleanly, plus eliminates needless server traffic.

#### Checklist

- [x] Closes #1076

### DIFF
--- a/tests/test_catalog.py
+++ b/tests/test_catalog.py
@@ -5,6 +5,7 @@
 
 from datetime import datetime
 import logging
+from unittest.mock import patch
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


lesserwhirls (approved):
Looks good to me!

